PR #6, opened 2014-03-21
Feature/get node properties

albertshau:


server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+      properties = getSet(body, "properties");
+      requiredServices = getSet(body, "services");
+    } catch (IllegalArgumentException e) {
+      responder.sendError(HttpResponseStatus.BAD_REQUEST, e.getMessage());
+      return;
+    }
+
+    JsonObject output = new JsonObject();
+    Set<Node> clusterNodes = clusterService.getClusterNodes(clusterId, userId);
+    if (clusterNodes.isEmpty()) {
```
sibonli: Is this statement necessary? If it is empty then it should ignore the for each loop and send the same response at the end.

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+   * is a JSON object with node ids as keys and JSON objects as values, where the value contains the properties passed
+   * in, or all properties if none were passed in.
+   *
+   * @param request Request for node properties in a cluster.
+   * @param responder Responder for sending the response.
+   * @throws Exception
+   */
+  @POST
+  @Path("/getNodeProperties")
+  public void getClusterNodes(HttpRequest request, HttpResponder responder) throws Exception {
```
sibonli: Would getNodeProperties() be a better name for this method?

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
```diff
@@ -700,6 +700,35 @@ public void deleteNode(String nodeId) throws Exception {
...
+        "SELECT N.node FROM nodes N, clusters C WHERE C.id=? AND C.owner_id=? AND N.cluster_id=C.id");
+      statement.setLong(1, Long.valueOf(clusterId));
+      statement.setString(2, userId);
+      try {
+        ResultSet rs = statement.executeQuery();
+        try {
+          while (rs.next()) {
+            Blob blob = rs.getBlob(1);
+            Node node = deserializeBlob(blob, Node.class);
+            nodes.add(node);
```
sibonli: This result deserialization block is seen in most of the methods in this class. Might be better create a separate method that is reused.

server/src/test/java/com/continuuity/loom/TestHelper.java
```diff
@@ -56,4 +59,20 @@ public static void finishTask(String loomUrl, JsonObject jsonObject) throws Exce
     Assert.assertEquals(200, response.getStatusLine().getStatusCode());
   }
 
+  public static JsonObject jsonObjectOf(String... keyvalues) {
+    Preconditions.checkArgument(keyvalues.length % 2 == 0, "must have a value for each key");
+    JsonObject output = new JsonObject();
+    for (int i = 0; i < keyvalues.length; i += 2) {
+      output.addProperty(keyvalues[i], keyvalues[i + 1]);
+    }
```
sibonli: Would be better if this method took in Map.Entries or some sort of K-V pair rather than a list of strings.

sibonli:
A few comments

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
albertshau: good catch, will remove it

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
albertshau: it would, will change

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+    String userId = getAndAuthenticateUser(request, responder);
+    if (userId == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+    JsonObject body;
+    try {
+      body = GSON.fromJson(reader, JsonObject.class);
+    } catch (Exception e) {
+      responder.sendError(HttpResponseStatus.BAD_REQUEST, "invalid request body. Must be a valid JSON Object.");
```
chtyim: invalid -> Invalid

albertshau: done

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+  @Path("/getNodeProperties")
+  public void getClusterNodes(HttpRequest request, HttpResponder responder) throws Exception {
+    String userId = getAndAuthenticateUser(request, responder);
+    if (userId == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+    JsonObject body;
+    try {
+      body = GSON.fromJson(reader, JsonObject.class);
```
chtyim: Why deserialize into JsonObject type? Why not use a class type that has the expected structure in it? E.g.

``` java
class ClusterNodesRequest {
  String clusterId;
  Set<String> properties;
  Set<String> services;
}
```

This would simplify a lot of the manual decoding logic here and make the code more reusable in future (say, you could encapsulate encode/decode logic into TypeAdapter if needed).

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+
+    for (Node node : clusterNodes) {
+      Set<String> nodeServices = Sets.newHashSet();
+      for (Service service : node.getServices()) {
+        nodeServices.add(service.getName());
+      }
+
+      // if the node has all services needed
+      if (nodeServices.containsAll(requiredServices)) {
+        JsonObject outputProperties;
```
chtyim, line 150: Try not to use Json type as data structure. E.g. this one is a `Map<String, String>`, isn't it? Using JsonObject everywhere makes the code highly difficult to read.

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
```diff
@@ -90,4 +101,105 @@ public void getClusterStatuses(HttpRequest request, HttpResponder responder) thr
...
+          outputProperties = node.getProperties();
+        }
+        output.add(node.getId(), outputProperties);
+      }
+    }
+
+    responder.sendJson(HttpResponseStatus.OK, output);
+  }
+
+  private Set<String> getSet(JsonObject object, String key) {
```
chtyim: We shouldn't be doing this kind of manual decode. See my comment above.

server/src/main/java/com/continuuity/loom/store/ClusterStore.java
```diff
@@ -189,6 +189,16 @@
...
 
   /**
+   * Get all nodes belonging to a specific cluster owned by a specific user.
+   * @param clusterId Id of the cluster whose nodes will be fetched.
+   * @param userId Id of the owner of the cluster.
+   * @return Set of all nodes belonging to specified cluster owned by the specified user.
+   *         Empty if no cluster or cluster nodes exist.
+   * @throws Exception if there was a problem getting the cluster nodes.
+   */
+  Set<Node> getClusterNodes(String clusterId, String userId) throws Exception;
```
chtyim, line 199: Is the returned Set mutable or immutable or unmodifiable (difference between the last two is immutable means it's will not change, unmodifiable means the content might get change internal, more like a view).

In general when returning Collection type object from a method, it's always important to state what kind of behavior of the instance has.

chtyim, line 199: Since this is an interface, it'd be easiest to say returned Set is immutable (as it impose less requirement on the implementation).

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
albertshau: good point, will do

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
albertshau, line 150: this one isn't a map of strings, the node properties can be populated by plugins further down and the only restriction we've set is that its a json object, so the values here can be any json element.

server/src/main/java/com/continuuity/loom/http/LoomRPCHandler.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
albertshau: agreed, will do

server/src/test/java/com/continuuity/loom/TestHelper.java
albertshau: this was purely a convenience method for testing so I'd rather not create objects to send in as that would defeat the purpose.  I might just remove this as its used in only one place for an object with 2 keyvalues

server/src/main/java/com/continuuity/loom/store/ClusterStore.java
albertshau, line 199: got it, will do

chtyim:
LGTM.

wolf31o2:
Resolves #4

PR #7, opened 2014-03-21
Feature/server set alternative

dereklwood:
- [x] setting higher update-alternatives priority for conf.loom directory to ensure it is made active
- [x] documenting current update-alternatives behavior in packaging


wolf31o2:
:+1: LGTM...

PR #8, opened 2014-03-21
removing dependency on /examples for ui integration tests, fix build

rns4731:


integration-testing/pom.xml
```diff
@@ -31,6 +31,24 @@
...
+    <repositories>
+        <repository>
+            <id>nexus</id>
+            <url>https://repository.continuuity.com/content/groups/public</url>
+        </repository>
+        <repository>
+            <id>releases</id>
+            <url>https://repository.continuuity.com/content/repositories/releases/</url>
+        </repository>
+        <repository>
```
albertshau, line 39: this repo is private and shouldn't be needed

integration-testing/README.md
```diff
@@ -2,5 +2,18 @@ Selenium testing for Loom.
 
 Runs with the test environment of loom ui server.
 
-node server.js --env=test
```
albertshau, line 5: do these tests require node.js? If so we should put a note here.  Any browser dependencies?

integration-testing/pom.xml
rns4731, line 39: done.

integration-testing/README.md
rns4731, line 5: done.

PR #13, opened 2014-03-24
Issue #11: make default services a subset of compatible services

rns4731:
- Ensures default hardwaretypes, imagetypes and services are a subset of compatible services in the UI dropdowns.


dereklwood:
:+1: lgtm

PR #15, opened 2014-03-25
Feature/changing loom port

rns4731:


integration-testing/src/test/java/com/continuuity/test/Constants.java
```diff
@@ -24,8 +24,9 @@
  *
  */
 public class Constants {
+
   // URI
-  public static final String ROOT_URL = "http://localhost:8100";
+  public static final String ROOT_URL = "http://localhost:56974";
```
albertshau: if we split out the port as a separate constant we can use the same constant below instead of specifying it in 2 places

integration-testing/src/test/java/com/continuuity/test/utilities/SuiteOrder.java
```diff
@@ -71,12 +71,12 @@
   public static void setUpClass() throws Exception {
     ProcessBuilder builder = new ProcessBuilder();
     builder.directory(new File("../ui"));
-    builder.command(new String[]{"/bin/sh","-c","node server.js --env=test"});
+    builder.command("/bin/sh","-c","node server.js --env=test --port=56974");
```
albertshau: is there a way to make this configurable in case this port is taken as well?

integration-testing/src/test/java/com/continuuity/test/Constants.java
rns4731: done.

integration-testing/src/test/java/com/continuuity/test/utilities/SuiteOrder.java
rns4731: done.

integration-testing/src/test/java/com/continuuity/test/Constants.java
```diff
@@ -93,4 +96,16 @@
   public static final String CLUSTER_CREATE_FILE_NAME = PARENT_PATH + "/ui/test/clusters/createcluster.json";
 
   public static final String ACTION_TABLE_CLASSNAME = ".node-actions-table";
+
+  public static int getPort() {
+    int port = -1;
+    try {
+      ServerSocket s = new ServerSocket(0);
```
albertshau, line 103: there is a chance that this port gets used between now and when node is started up.  If there is a way to tell node to bind to a random port and a way of getting that random port, that would be the best solution.  If not, this is good enough.

albertshau:
one comment, but if its not possible then lgtm

PR #26, opened 2014-03-28
Changing port finder to dynamically allocate port

rns4731:


integration-testing/src/test/java/com/continuuity/test/Constants.java
```diff
@@ -97,15 +99,44 @@
...
+    throw new RuntimeException("Could not find an available port between " +
+                                 MIN_PORT_NUMBER + " and " + MAX_PORT_NUMBER);
+  }
+
+  private static boolean available(final int port) {
+    ServerSocket serverSocket = null;
+    DatagramSocket dataSocket = null;
     try {
-      ServerSocket s = new ServerSocket(0);
-      port = s.getLocalPort();
```
albertshau, line 104: just need a s.close() here instead of all the other changes

PR #36, opened 2014-04-02
Fixed a few more spots

sibonli:


albertshau:
If I grep for loom I see several more places where it's missing Continuuity:

grep -i loom source -r | grep -v Continuuity | grep -v loom-server | grep -v loom-provisioner | grep -v 'X-Loom-'

Can we change those places as well

albertshau:
lgtm

PR #42, opened 2014-04-04
Feature/shell provisioner json

dereklwood:
This PR provides a utility function to the shell provisioner to lookup values in the task json.  This makes use of https://github.com/rcrowley/json.sh, for which the license is included
- [x] shell provisioner modified to generate/scp the task json data into remote cache and call a wrapper script instead of user scripts directly
- [x] wrapper script to invoke json.sh on the current task data and provide loom_lookup_key

example, for service:

```
                "actions": {
                    "install": {
                        "type": "shell",
                        "script": "examples/json_substitution_example.sh",
                        "data": "taskId config/automators"
                    }
                }
```

results:

```
2014-04-03 18:29:34 -0700 DEBUG: Result of shell command: {"stdout"=>"--- example json parsing ---\njson lookup: taskId --> 00000011-030-076\njson lookup: config/automators --> chef shell\n", "stderr"=>"", "status"=>0}
```


provisioner/daemon/plugins/automators/shell_automator/scripts/.lib/loom_wrapper.sh
```diff
@@ -0,0 +1,66 @@
...
+      else
+        local _notanarray="true"
+        break
+      fi
+    done <<< "${_matches}"
+    if [ "x$_notanarray" != "x" ]; then
+      echo "non-unique key, multiple matches found" 1>&2;
+      return 1
+    else
+      # success; return space-delimited array values
```
albertshau, line 54: what happens if there are spaces in the values? Are the values wrapped in quotes?

albertshau, line 54: looks like they are

albertshau:
lgtm though we should be sure to add documentation this either on this branch or separately.

provisioner/daemon/plugins/automators/shell_automator/scripts/.lib/loom_wrapper.sh
dereklwood, line 54: yes, spaces in values are fine.  spaces in keys are not

PR #44, opened 2014-04-04
selective bootstrapping

dereklwood:
Provisioner will now run only the bootstrap tasks for the plugins specified in `{"config": {"automators": [] } }`.  

If json not present or empty, default to current behavior of running all plugin bootstrap tasks.

### DIFF
--- a/provisioner/daemon/provisioner.rb
+++ b/provisioner/daemon/provisioner.rb
@@ -98,16 +98,27 @@ def delegate_task(task, pluginmanager)
     object = clazz.new(task)
     result = object.runTask
   when 'bootstrap'
-    # for bootstrap we run all registered bootstrap handlers
-    classes = pluginmanager.getAllHandlerActionObjectsForAutomators()
-    raise "No bootstrappers configured" if classes.empty?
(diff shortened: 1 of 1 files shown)


provisioner/daemon/provisioner.rb
```diff
@@ -98,16 +98,27 @@ def delegate_task(task, pluginmanager)
...
+        object = clazz.new(task)
+        result = object.runTask
+        combinedresult.merge!(result)
+      end
+    else
+      # default to running all registered bootstrap handlers
+      classes = pluginmanager.getAllHandlerActionObjectsForAutomators()
+      raise "No bootstrappers configured" if classes.empty?
+      log.debug "Task #{task_id} running bootstrap handlers: #{classes}"
+      classes.each do |clazz|
```
albertshau: we can run this loop after the entire 'if - else' instead of in both clauses if we set the classes array in the if condition.

dereklwood: fixed

albertshau:
lgtm

wolf31o2:
:+1: LGTM

PR #45, opened 2014-04-04
Set YARN NodeManager memory constraints to half the machine's total RAM

wolf31o2:
YARN NodeManager will try to use all of the available RAM on a system, even if it is allocated. Setting this option puts a sane upper bound on YARN memory usage.


albertshau:
lgtm

PR #46, opened 2014-04-04
Update loom-defaults.sh LOOM_SERVER_URI env var

wolf31o2:
Make LOOM_SERVER_URI in load-defaults.sh match to format of LOOM_SERVER_URI used in start scripts


nitinmotgi:
lgtm :+1:

PR #53, opened 2014-04-05
Feature/remove omnibus

wolf31o2:
All deletes, no adds.

The omnibus configuration and build for Continuuity Loom was confusing, undocumented, and has a ton of requirements. As such, we're removing it from the project.


albertshau:
lgtm

PR #54, opened 2014-04-05
Feature/update rackspace joyent gems

dereklwood:
updating rackspace/joyent provider plugin gem dependencies to their latest corresponding knife-plugin versions.
- rackspace requires a new "region" option which is now surfaced in the UI
- no significant changes in joyent, just syncing the relevant create/confirm code blocks
- updated example configs, docs


provisioner/daemon/plugins/providers/rackspace_provider/chef/knife/loom_rackspace_server_create.rb
```diff
@@ -74,6 +76,7 @@ def run
         msg_pair("RackConnect Wait", rackconnect_wait ? 'yes' : 'no')
         msg_pair("ServiceLevel Wait", rackspace_servicelevel_wait ? 'yes' : 'no')
         msg_pair("Password", server.password)
+        msg_pair("SSH Key", Chef::Config[:knife][:rackspace_ssh_keypair])
```
wolf31o2, line 79: Should we add if to the end of this to keep it from displaying, when we don't use a key pair?

dereklwood, line 79: msg_pair() will do this already I believe

wolf31o2:
:+1: LGTM

PR #57, opened 2014-04-09
Feature/change paths

wolf31o2:
- [x] Update PATHs to not include component when looking for embedded binaries and libraries
- [x] Update config/defaults PATH for standalone to match non-standalone locations


albertshau:
lgtm

PR #59, opened 2014-04-09
Import ntp version 1.6.2

wolf31o2:
Pulling this in, so that we can use it, where necessary.


wolf31o2:
Ignore this one. This should be going to release/0.9.6 branch.

PR #60, opened 2014-04-09
Import ntp version 1.6.2

wolf31o2:
This is useful for clock sync, which is necessary for many coordinated services.


poornachandra:
LGTM :+1:

PR #64, opened 2014-04-11
adding ability to reconfigure an active cluster.

albertshau:


server/src/main/java/com/continuuity/loom/scheduler/Actions.java
```diff
@@ -115,7 +134,14 @@ public Actions() {
                         // Start of a dependent service can happen only after its dependency has started
                         new Dependency(ProvisionerAction.START, ProvisionerAction.START),
                         // Initialize of a dependent service can happen only after its dependency has started
-                        new Dependency(ProvisionerAction.START, ProvisionerAction.INITIALIZE))
+                        new Dependency(ProvisionerAction.START, ProvisionerAction.INITIALIZE)),
+        ClusterAction.CLUSTER_CONFIGURE_WITH_RESTART,
+                      ImmutableSet.of(
+                        // Start of a dependent service can happen only after its dependency has started
+                        new Dependency(ProvisionerAction.START, ProvisionerAction.START),
+                        // Stop of a dependent service can happen only after its dependency has started
```
poornachandra: "after its dependency has started" -> "after its dependency has stopped"

server/src/main/java/com/continuuity/loom/scheduler/task/ClusterService.java
```diff
@@ -39,6 +41,8 @@
  */
 public class ClusterService {
   private static final Logger LOG = LoggerFactory.getLogger(ClusterService.class);
+  private static final ImmutableSet<Cluster.Status> CONFIGURABLE_STATES =
```
poornachandra: Maybe move this to the file where Cluster Status is defined.

chtyim: Use more generic interface, such as `Set`, rather than a specific implementation as type.

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
```diff
@@ -177,10 +177,37 @@ public Cluster getCluster(String clusterId, String ownerId) throws Exception {
...
+      conn.close();
+    }
+  }
+
+  @Override
   public void writeCluster(Cluster cluster) throws Exception {
     // sticking with standard sql... this could be done in one step with replace, or with
     // insert ... on duplicate key update with mysql.
-    long clusterNum = Long.valueOf(cluster.getId());
+    long clusterNum = Long.parseLong(cluster.getId());
```
poornachandra, line 234: Just curious - why is valueOf changed to parseLong?

poornachandra:
Minor comments, otherwise LGTM.

I haven't reviewed the html/css/js files though. I assume they are result of a merge.

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
```diff
@@ -714,6 +741,26 @@ public void deleteNode(String nodeId) throws Exception {
...
+  private boolean hasResults(PreparedStatement statement) throws SQLException {
+    try {
+      ResultSet rs = statement.executeQuery();
+      try {
+        return rs.next();
+      } finally {
+        rs.close();
+      }
+    } finally {
+      statement.close();
```
chtyim: Not a good practice to close a resource being passed to a method (except a util method for performing the close, e.g. in `Closeables.closeQuietly()`). Resource acquisition and release should be in the same block (in this case, caller should be closing the statement).

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
albertshau, line 234: feedback from a previous PR, parseLong returns a long whereas valueOf returns the Long object so better to use parseLong since the left is a long.

server/src/main/java/com/continuuity/loom/scheduler/task/ClusterService.java
albertshau: done for both

server/src/main/java/com/continuuity/loom/scheduler/Actions.java
albertshau: done.  It should actually be "before its dependency has stopped" too.

server/src/main/java/com/continuuity/loom/store/SQLClusterStore.java
albertshau: done

PR #66, opened 2014-04-14
adding ability to plan jobs for a delta of the services and/or

albertshau:
nodes in a cluster to support cluster operations that act on an
existing cluster. Additionally, refactored most of the job
planning logic from the ClusterScheduler into a new JobPlanner
class.


server/src/main/java/com/continuuity/loom/scheduler/JobPlanner.java
```diff
@@ -0,0 +1,315 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+
+/**
+ *
```
poornachandra: Missing javadoc!

poornachandra:
The only major change is adding "should plan node" and "should plan service" to JobPlanner, right? If so changes LGTM.

It is a bit difficult to review the code changes due to the massive refactoring. Maybe next time do the code change and refactoring in separate PRs.

albertshau:
that's correct, should plan node and should plan service are the logic changes.  Next time I'll do the refactor in a separate PR.

PR #70, opened 2014-04-14
Add install action to base service

wolf31o2:
The base service may install packages, such as iptables, so it should have an install action.


albertshau:
lgtm

PR #74, opened 2014-04-15
restructuring poms to link server module to integration-testing,

albertshau:
consolidating versions and common dependencies.

### DIFF
--- a/integration-testing/pom.xml
+++ b/integration-testing/pom.xml
@@ -21,38 +21,16 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>com.continuuity</groupId>
     <artifactId>integration-testing</artifactId>
     <packaging>jar</packaging>
-    <version>0.9.6</version>
     <name>Loom UI testing</name>
-
-    <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
-    </properties>
-    <repositories>
-        <repository>
-            <id>nexus</id>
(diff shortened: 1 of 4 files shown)


integration-testing/pom.xml
```diff
@@ -21,39 +21,17 @@
...
-        </repository>
-        <repository>
-            <id>continuuity-release</id>
-            <url>https://repository.continuuity.com/content/repositories/releases-public/</url>
-        </repository>
-    </repositories>
+    <parent>
+        <groupId>com.continuuity</groupId>
+        <artifactId>loom-parent</artifactId>
+        <version>0.9.7-SNAPSHOT</version>
```
wolf31o2, line 30: 0.9.7-SNAPSHOT or 0.9.6-SNAPSHOT?

I am guessing the former is supposed to mean 0.9.7-somelargenumberhere and is supposed to be the equivalent of 0.9.7_pre?

albertshau, line 30: Correct, 0.9.7-SNAPSHOT means it's what will eventually become 0.9.7.

pom.xml
```diff
@@ -0,0 +1,267 @@
...
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-javadoc-plugin</artifactId>
+                    <version>2.9</version>
+                    <configuration>
+                        <excludePackageNames>*.internal.*</excludePackageNames>
+                        <links>
+                            <link>http://download.oracle.com/javase/6/docs/api/</link>
+                        </links>
+                        <bottom>
+                            <![CDATA[Copyright 2013 Continuuity, Inc. <a href="http://www.apache.org/licenses/LICENSE-2.0">Licensed under the Apache License, Version 2.0</a>]]>
```
wolf31o2: 2013-2014

albertshau: good catch

wolf31o2:
Looks good. I really dislike the top-level pom.xml, since it isn't for the entire project, but that shouldn't block merge.

PR #78, opened 2014-04-16
refactoring solver code to be more generic with the introduction

albertshau:
of a ClusterLayoutChange, which can apply some sort of change
to a ClusterLayout. Currently only an AddServicesChange is
implemented, but we can add ones for removing a service or adding
nodes in the future. Also changing the solver interface a bit
to return a diff of cluster layout when adding services to an
existing layout.

### DIFF
--- a/server/src/main/java/com/continuuity/loom/layout/change/ClusterLayoutChange.java
+++ b/server/src/main/java/com/continuuity/loom/layout/change/ClusterLayoutChange.java
@@ -18,7 +18,10 @@ import java.util.Set;
 public interface ClusterLayoutChange {
 
   /**
-   * Returns whether or not the change can be applied to the given {@link ClusterLayout}.
+   * Returns whether or not the change can be applied to the given {@link ClusterLayout}. This is not for checking
+   * if the resulting cluster layout would satisfy all its constraints, but for checking whether or not it is physically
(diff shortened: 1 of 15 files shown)


server/src/main/java/com/continuuity/loom/layout/change/ClusterLayoutChange.java
```diff
@@ -0,0 +1,46 @@
...
+ * Represents some change to a {@link ClusterLayout}, such as adding a service to some of the nodes. Caller is
+ * responsible for calling the {@link #canApplyChange(com.continuuity.loom.layout.ClusterLayout)} method to validate
+ * if a change can be applied to a layout before actually applying the change with
+ * {@link #applyChange(com.continuuity.loom.layout.ClusterLayout)} or
+ * {@link #applyChange(com.continuuity.loom.cluster.Cluster, java.util.Set, java.util.Map)}.
+ */
+public interface ClusterLayoutChange {
+
+  /**
+   * Returns whether or not the change can be applied to the given {@link ClusterLayout}.
```
poornachandra: Can you add some details as to when a change can not be applied?

poornachandra:
One minor comment, otherwise LGTM :+1:

PR #83, opened 2014-04-18
Feature/add service apis

albertshau:


server/src/main/java/com/continuuity/loom/layout/ClusterLayout.java
```diff
@@ -63,6 +61,28 @@ public Constraints getConstraints() {
   }
 
   /**
+   * Determine a cluster layout based on a set of nodes and their constraints.
+   *
+   * @param clusterNodes Nodes of a cluster.
+   * @param constraints Constraints associated with the cluster.
+   * @return The ClusterLayout derived from the nodes of a cluster.
+   */
+  public static ClusterLayout fromClusterNodes(Set<Node> clusterNodes, Constraints constraints) {
```
poornachandra: What is the difference between fromClusterNodes and fromNodes?

server/src/main/java/com/continuuity/loom/scheduler/ClusterScheduler.java
```diff
@@ -141,7 +141,19 @@ public void run() {
...
+          switch (clusterAction) {
+            case CLUSTER_CREATE:
+              taskService.failJobAndTerminateCluster(job, cluster, "Failed to schedule the action");
+              break;
+            default:
+              // failed to plan means the job should fail, but since no actions were performed on the cluster
+              // it should stay in the same state as before.
+              // TODO: at this point the cluster state has been changed. Should revert it here but need versioning
+              //       or cluster history or something to that effect.
+              taskService.failJobAndSetClusterStatus(job, cluster, Cluster.Status.ACTIVE,
```
poornachandra: Since the cluster state has been changed but the job failed, shouldn't the cluster state go into INCONSISTENT instead of ACTIVE?

We can later give an option to re-run the job if the failure was temporary. Or we can think of clearing the state and then putting the cluster status to ACTIVE.

server/src/main/java/com/continuuity/loom/layout/ClusterLayout.java
albertshau: good catch this was a rebase error

server/src/main/java/com/continuuity/loom/scheduler/ClusterScheduler.java
albertshau: you're right it should be inconsistent until we have rollback.

PR #84, opened 2014-04-21
Feature/update mysql cookbook

wolf31o2:
Pull in version 4.1.2 of the mysql cookbook and its dependencies. Versions after 4.1.2 change how the cookbook works and need much more testing before we can merge it.


wolf31o2:
Over the shoulder approval from @dereklwood since this is just pulling in upstream cookbooks

PR #90, opened 2014-04-23
changing default admin password to something easier since its not

albertshau:
real authentication yet anyway.


wolf31o2:
:+1: LGTM

PR #93, opened 2014-04-24
add callbacks before and after cluster operations. Refactoring

albertshau:
some common logic into TaskService to make sure the before and
after hooks are isolated to a single class.


server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.util.Iterator;
+import java.util.Set;
+
+/**
+ * Executes before and after hooks by sending an HTTP POST request to some configurable endpoints, with the post body
+ * containing the cluster and job objects. By default the requests are sent before and after cluster creation and
+ * cluster deletion,
```
nitinmotgi: Is there cluster creation failure, cluster update and service start/stop ?

albertshau: yes, these get called before and after (failure counts as after) any cluster operation including service stop,start,restart and reconfigure.  This particular callback I've defaulted to only do on create and delete.  Do you think it should default to every cluster operation?

albertshau: hm maybe it would be better to have another method for failures instead of lumping it into the after call

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private void sendPost(String url, CallbackData data) {
+    if (url != null) {
+      HttpPost post = new HttpPost(url);
+
+      try {
+        post.setEntity(new StringEntity(GSON.toJson(data)));
+        HttpResponse response = httpClient.execute(post);
+        // ignore what we get back, but make sure resources are released
+        HttpEntity entity = response.getEntity();
```
poornachandra: This call needs to be async so that the pipeline does not get delayed.

server/src/main/java/com/continuuity/loom/scheduler/callback/ClusterCallbackExecutor.java
```diff
@@ -0,0 +1,101 @@
...
+   * @param data Data available to use while executing callback.
+   */
+  public void onStart(final CallbackData data) {
+    ListenableFuture future = executorService.submit(new Runnable() {
+      @Override
+      public void run() {
+        clusterCallback.onStart(data);
+      }
+    });
+    future.addListener(new Runnable() {
```
poornachandra: When the server goes down after calling the callback and comes back up, the future is lost and the job will not make any progress, right?

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -0,0 +1,167 @@
...
+        if (entity != null) {
+          entity.getContent().close();
```
poornachandra: It is generally a good practice to have close in a finally block.

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+    if (onFailureUrl != null) {
+      LOG.debug("after hook will be triggered on actions {}", Joiner.on(',').join(failureTriggerActions));
+    }
+
+    int maxConnections = conf.getInt(Constants.HttpCallback.MAX_CONNECTIONS,
+                                     Constants.HttpCallback.DEFAULT_MAX_CONNECTIONS);
+    PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
+    connectionManager.setDefaultMaxPerRoute(maxConnections);
+    connectionManager.setMaxTotal(maxConnections);
```
poornachandra, line 93: What happens when maxConnections is reached?

server/src/test/java/com/continuuity/loom/scheduler/SchedulerTest.java
```diff
@@ -222,6 +227,65 @@ public void testScheduler() throws Exception {
...
+    clusterStore.writeNode(node);
+
+    node = GSON.fromJson(NODE2, Node.class);
+    clusterStore.writeNode(node);
+
+    inputQueue.add(new Element(cluster.getId(), ClusterAction.CLUSTER_CREATE.name()));
+    clusterScheduler.run();
+
+    // at this point, the start callback should have run, but not the after callbacks
+    Assert.assertEquals(1, mockClusterCallback.getStartCallbacks().size());
```
poornachandra, line 285: Since you mentioned that until the onStart callback has returned the job should not start, I think it is a good idea to assert that.

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Set;
+
+/**
+ * Executes before and after hooks by sending an HTTP POST request to some configurable endpoints, with the post body
+ * containing the cluster and job objects, assuming there is a valid url assigned to the start, success, and/or failure
+ * urls. If no url is specified, no request will be sent. Additionally, trigger actions can be configured so that
+ * the HTTP POST request is sent only for specific cluster actions. This is done by specifying a comma separated list
+ * of {@link ClusterAction}s in the configuration for start, success, and/or triggers.
+ */
+public class HttpPostClusterCallback implements ClusterCallback {
```
poornachandra, line 56: It would be good to have tests for this class.

server/src/main/java/com/continuuity/loom/scheduler/JobScheduler.java
```diff
@@ -99,14 +95,27 @@ public void run() {
...
           ClusterJob job = clusterStore.getClusterJob(jobId);
+          // TODO: avoid queueing job when its already complete
+          if (job.getJobStatus() == ClusterJob.Status.COMPLETE) {
+            continue;
+          }
+          Cluster cluster = clusterStore.getCluster(job.getClusterId());
+          // this can happen if 2 tasks complete around the same time and the first one places the job in the queue,
+          // sees 0 in progress tasks, and sets the cluster status. The job is still in the queue as another element
+          // from the 2nd task and gets here.  In that case, no need to go further.
+          if (job.getJobStatus() == ClusterJob.Status.FAILED && cluster.getStatus() != Cluster.Status.PENDING) {
```
poornachandra: What about setting the task status? Same comment for COMPLETE above.

albertshau: the task itself has already been updated at this point so there is no need.  We add the job onto the queue after the task status is marked as complete in order to check if the stage needs to be advanced of the job itself is done.

server/src/test/java/com/continuuity/loom/scheduler/SchedulerTest.java
albertshau, line 285: done

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
albertshau, line 56: done

server/src/main/java/com/continuuity/loom/scheduler/callback/ClusterCallbackExecutor.java
albertshau: will use a queue like in other places for this.  Also means the callback should be idempotent, which will be documented

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
albertshau, line 93: blocks on the post call until connections go down.

server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
albertshau: this actually isn't needed after upgrading the client version and having post.releaseConnection in the finally.  will take it out.

poornachandra:
LGTM :+1:

PR #107, opened 2014-04-28
Fix aliases in /etc/hosts for machines using a DNS suffix other than '.local'

wolf31o2:


albertshau:
lgtm

albertshau:
is it easy to add a unit test for this?

wolf31o2:
I can write a test-suite for this cookbook. It wouldn't be too hard, but it wouldn't be something that gets tested, automatically.

wolf31o2:
Remember, we still don't have unit tests on the provisioner. Many individual cookbooks with their own repositories have tests and are tested when things change in those repos, but there's not a global one for the provisioner (yet).

PR #114, opened 2014-04-30
Update imagetypes for centos6 and ubuntu12 to latest from Joyent/Rackspace

wolf31o2:
These images include fixes for Heartbleed


dereklwood:
:+1: lgtm

per @wolf31o2, not using rackspace's PVHVM image without adequate testing.

PR #115, opened 2014-05-01
Feature/dotchef knifedotrb

wolf31o2:
Adding knife.rb to chef_automator to make it easier to work with Chef cookbooks. Otherwise, you must always specify the cookbook_path when running knife commands, such as `knife cookbook test` or `knife cookbook site install`...


dereklwood:
:+1: lgtm

PR #118, opened 2014-05-02
Hotfix/fixes for latest images

dereklwood:
updating to the latest provider images raised several issues, which this PR addresses:
- [x] curl not included in the minimal images.  for chef install, added a check for curl and default to wget if not installed
- [x] scp not installed in the minimal images.  Added a check for it, and attempt a yum or apt-get install.  This is a bit ugly, and needs to be added to both automators.
- [x] rackspace new images empirically more susceptible to connection refused errors.  added the tcp_test_ssh and sleep from the latest knife_rackspace gem


wolf31o2:
:+1: Do eet!

PR #120, opened 2014-05-02
fix debug statement to use correct variable

dereklwood:
fixes bug in plugin registration


wolf31o2:
:+1: LGTM... and merging

PR #131, opened 2014-05-07
Feature/import nginx cookbook

wolf31o2:
This is an import of the nginx cookbook and dependencies. Up next, services and clustertemplates!


dereklwood:
all 3rd party code, nothing to review

PR #133, opened 2014-05-07
Feature/rspec

wolf31o2:
This is the beginning of a testing framework. Test coverage is about 21% for `automator.rb`, `pluginmanager.rb`, `provider.rb`, and `utils.rb` files. There's no coverage of `provisioner.rb` since it implements no classes that can be easily tested. This is using SimpleCov to create coverage reports.


provisioner/spec/automator_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'json'
+
+describe Automator do
+
+  response = IO.read("#{File.dirname(__FILE__)}/task.json")
+
+  # Set these up once
+  before :all do
+    %w(bootstrap install configure initialize start stop remove).each do |taskname|
+      instance_variable_set("@task_#{taskname}", JSON.parse(response.to_str.gsub('BOOTSTRAP', taskname)))
```
dereklwood: to_str is redundant here

provisioner/spec/automator_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+        it "creates task instance variable" do
+          expect(instance_variable_get("@automator_#{taskname}").task).to eql instance_variable_get("@task_#{taskname}")
+        end
+        it "creates empty result hash" do
+          expect(instance_variable_get("@automator_#{taskname}").result).to be_empty
+        end
+      end
+
+# TODO:
```
dereklwood: fails rubocop:

```
spec/automator_spec.rb:31:3: C: Annotation keywords should be all upper case, followed by a colon and a space, then a note describing the problem.
# TODO:
  ^^^^^
```

dereklwood:
couple comments, otherwise LGTM

provisioner/spec/automator_spec.rb
wolf31o2: IO.read a file makes it a string?

PR #134, opened 2014-05-08
Import hadoop version 1.0.8

wolf31o2:
This is a simple cookbook import. This is needed for HDP 2.1 updates, which is why it is going to release/0.9.7 and not to develop.


dereklwood:
:+1:

PR #136, opened 2014-05-08
Hotfix/fix install doc

dereklwood:
a few doc fixes I encountered when setting up a new server (0.9.6, but seems these impact 0.9.7 as well)


wolf31o2:
:+1: LGTM

PR #138, opened 2014-05-08
Feature/latest job compatibility

albertshau:


wolf31o2:
Well, it looks alright to me, but we should definitely have someone more familiar with Java check it out, too.

albertshau:
poorna says lgtm

PR #139, opened 2014-05-09
Hotfix/shell provisioner invocation

dereklwood:
This fixes a bug in the current shell automator where it fails if you specify anything other than an included script.  It was attempting to source whatever you gave it.

This PR attempts to detect if it's a bash script and only then will source it.  Otherwise, it just executes it.

### DIFF
--- a/provisioner/daemon/plugins/automators/shell_automator/scripts/.lib/loom_wrapper.sh
+++ b/provisioner/daemon/plugins/automators/shell_automator/scripts/.lib/loom_wrapper.sh
@@ -70,11 +70,15 @@ if [ -f ${USERSCRIPT} ]; then
     # source it, so that it can make use of the function(s) above
     . ${USERSCRIPT} $*
   else
-    # not a shell script, just execute it
-    ./${USERSCRIPT} $*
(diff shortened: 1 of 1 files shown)


provisioner/daemon/plugins/automators/shell_automator/scripts/.lib/loom_wrapper.sh
```diff
@@ -63,5 +63,18 @@ loom_lookup_key () {
...
-. $USERSCRIPT $*
+if [ -f ${USERSCRIPT} ]; then
+  # USERSCRIPT is a file.  Check if it is a shell script
+  file ${USERSCRIPT} | grep "shell.*executable" 2>&1 >/dev/null
+  if [ $? == 0 ]; then
+    # source it, so that it can make use of the function(s) above
+    . ${USERSCRIPT} $*
+  else
+    # not a shell script, just execute it
+    ./${USERSCRIPT} $*
```
wolf31o2: What if it isn't executable? Just let it fail?

wolf31o2:
:+1: LGTM

PR #146, opened 2014-05-12
Feature/add hive services

wolf31o2:
This adds services for `hive-metastore` and `hive-server2` with an optional `hive-metastore-database` to configure `mysql-server` for Hive. Since we now have extended dependency support and the ability to add services to clusters, this also gets rid of the `hadoop-hbase-distributed` and `hadoop-hbase-singlenode` templates and rolls their configuration into `hadoop-distributed` and `hadoop-singlenode`, respectively. This would have otherwise required creation of templates for every permutation of services.


config/defaults/clustertemplates/hadoop-singlenode.json
```diff
@@ -47,13 +72,32 @@
...
+          "hadoop-yarn-nodemanager",
+          "hbase-regionserver",
+          "hadoop-hdfs-namenode",
+          "hadoop-yarn-resourcemanager",
+          "zookeeper-server",
+          "hbase-master",
+          "mysql-server",
+          "hive-metastore-database",
+          "hive-metastore",
+          "hive-server2"
```
albertshau, line 100: If we're being totally rigorous, this template doesn't force a 1 node cluster unless "base" is in this list, since somebody will be able to ask for more machines, and it will be valid to create nodes with just "base" on them.

albertshau, line 100: I don't think it's that big of a deal though since the real solution to this is to add a cluster size constraint instead of forcing strange must coexist constraints.

config/defaults/clustertemplates/hadoop-distributed.json
```diff
@@ -74,6 +122,38 @@
...
+        "quantities": {
+          "min": "1",
+          "max": "1"
+        }
+      },
+      "mysql-server": {
+        "quantities": {
+          "min": 1,
+          "max": 1,
+          "stepSize": 1
```
albertshau: should remove stepSize since it's not actually used

albertshau:
lgtm after removing stepSize

PR #147, opened 2014-05-12
Feature/improve service set performance

albertshau:
this should fix performance problems we were seeing with templates that had 10+ constrained services.


poornachandra:
LGTM :+1:

PR #149, opened 2014-05-13
Feature/move base dep to install

wolf31o2:
- [x] Switch to new dependency system for all services.
- [x] Move "base" to install/requires

Fixes #1 and fixes #96


albertshau:
lgtm

PR #151, opened 2014-05-13
Feature/merge release

wolf31o2:
This merges current release/0.9.7 back to develop, so that we can easily add features to either, as needed.


albertshau:
lgtm

PR #157, opened 2014-05-14
constrain to single hive-metastore

dereklwood:
constraining hive-metastore service to a single instance per cloudera docs, etc.


wolf31o2:
:+1: LGTM

PR #163, opened 2014-05-14
Feature/dont log when callback doesnt run

albertshau:


server/src/main/java/com/continuuity/loom/scheduler/callback/HttpPostClusterCallback.java
```diff
@@ -142,41 +146,41 @@ public void onSuccess(CallbackData data) {
...
+      if (onFailureUrl != null) {
+        LOG.debug("{} failed on cluster {}, sending request to {}",
+                  jobAction, data.getCluster().getId(), onFailureUrl);
+        sendPost(onFailureUrl, data);
+      }
     }
   }
 
   private void sendPost(String url, CallbackData data) {
-    if (url != null) {
```
wolf31o2, line 152: Why the if removal here? (Trying to get my learn on...)

albertshau, line 152: it's not wrong to keep it here, but it would be redundant since the null check is now done by callers of the function.

wolf31o2, line 152: That's what I thought. Thanks...

wolf31o2:
:+1: LGTM

PR #175, opened 2014-05-15
Hotfix/kerberos support

wolf31o2:
This fixes the `kerberos-client` and `kerberos-master` service JSON to use the correct fields for `chef-solo` automator. This also adds `kerberos-client` to install/uses and `kerberos-master` to runtime/uses on hadoop services (for secure Hadoop)...


dereklwood:
:+1: lgtm

PR #180, opened 2014-05-16
Feature/hive zookeeper quorum fix

wolf31o2:
Removes port


albertshau:
does this mean port is a separate config setting?

in any case, lgtm

wolf31o2:
It is, but it defaults to 2181, so we don't need to set it.

PR #183, opened 2014-05-16
Hotfix/kerberos client

wolf31o2:
This fixes `kerberos-client` by moving its runtime dependency on `kerberos-master` to uses instead of requires.


albertshau:
lgtm

PR #188, opened 2014-05-20
Feature/ui bugfixes

rns4731:
- Remove modification ability when cluster is deleted.
- Change links to docs and support.


wolf31o2:
:+1:

PR #191, opened 2014-05-20
Feature/put in missing commit

albertshau:
was part of an earlier merged PR that I forgot to push.


albertshau:
was part of issue 189 but forgot to push, merging.

PR #197, opened 2014-05-22
Hotfix/remove kerberos singlenode

wolf31o2:
This is not generic and needs to be removed. The services work, when used in another template. This template is misconfigured and should not have been included.


albertshau:
lgtm

PR #199, opened 2014-05-22
Hotfix/give yarn more ram

wolf31o2:
This gives YARN a little bit more RAM, which helps on smaller memory machines.


albertshau:
does this change depending on the OS or will this always be correct? in any case, lgtm.

wolf31o2:
It should always be the same, but it's really up to the kernel to provide the information, and chef/ohai to get it. On Linux, it should always be the same.

PR #200, opened 2014-05-22
Hotfix/yarn container executor

wolf31o2:
This option was causing glibc realloc errors. We need to investigate further before turning it back on...


albertshau:
lgtm

PR #205, opened 2014-05-23
Feature/upgrade guide

wolf31o2:
Add upgrade guide for Loom package users...


docs/docs/source/overview/upgrade-guide.rst
```diff
@@ -0,0 +1,112 @@
...
+  .. parsed-literal::
+   $ sudo /etc/init.d/loom-provisioner stop
+   $ sudo /etc/init.d/loom-ui stop
+   $ sudo /etc/init.d/loom-server stop
+
+* 3 - Stop Database and take a backup
+
+  * The embedded Derby database and ZooKeeper server are stopped when Loom Server is stopped
+  * If you are using another database, such as MySQL, follow your standard procedures for backing up the Loom database
+  * Replace ``/var/loom/data`` below with the location from ``loom.local.data.dir`` in ``/etc/loom/conf/loom-site.xml`` if you have modified it from default
```
albertshau, line 51: this line is only needed if the embedded derby db is being used, or if the in-process zookeeper server is being used.  I think it would be clearer to put 2 lines with 'if you are using the embedded derby db...' and 'if you are using the in-process zookeeper server...'

albertshau:
one comment, otherwise lgtm

docs/docs/source/overview/upgrade-guide.rst
wolf31o2, line 51: Updated. Good?

dereklwood, line 51: I think the \* Replace... line now should not be its own bullet point, but in parenthesis in the line above

PR #220, opened 2014-05-28
Fixing tests that were not revisited when callbacks were moved

albertshau:
to run in their own threads. Also removing print statements.


poornachandra:
LGTM :+1:

PR #224, opened 2014-05-30
updating configuration settings in example and put in a skeleton

albertshau:
upgrade guide that should be updated as things change.


wolf31o2:
IMO, the upgrade guide should be additive, meaning you shouldn't replace the 0.9.7 instructions. Rather, we should add the 0.9.8 instructions.

albertshau:
will do this in a separate PR

PR #225, opened 2014-06-02
adding ability to use a dummy provisioner for demo or debug

albertshau:
purposes.


albertshau:
not sure if this should be documented

wolf31o2:
:+1: LGTM... I don't think that we need to document this

PR #228, opened 2014-06-02
add default automator types for the dummy provisioner to register

albertshau:
with so that services can be viewed and edited when the dummy
provisioner is used.


albertshau:
also not documenting this. Should be temporary until plugins are managed by the server.

dereklwood:
:+1: lgtm

PR #229, opened 2014-06-02
Feature/fix failure output

dereklwood:
Making the ssh_exec error handling a little less ugly.  At a high level, it used to:
- [ ] plugin logs context, invokes ssh_exec, examines returned output to log more context and re-raise (which is buggy since CommandExecutionError wasn't caught)

Now, it:
- [x] enhances ssh_exec and CommandExecutionError to carry context fields, so plugin layer no longer needs to try to examine output and instead simply invoke with context and just let exceptions raise to top layer


wolf31o2:
:+1: LGTM

PR #237, opened 2014-06-07
Feature/memcached

wolf31o2:
This adds `memcached` support to `lamp` clustertemplate... all dependencies were already present. :-)


wolf31o2:
This had a php-memcache service, which has been removed. This will be added back later. I chose this to keep from using deprecated support in the php cookbook.

wolf31o2:
This can be added back to `lamp` after #317 is resolved

albertshau:
lgtm

PR #238, opened 2014-06-09
Feature/apache modules perl php python

wolf31o2:
This updates LAMP to support Perl and Python. It also adds the PHP5 module, to fix PHP support. Previously, PHP would be installed, but the web server did not know how to serve up PHP pages.

The list of modules is taken from the defaults in the apache2 cookbook, with perl, python, php5, and ssl added.


dereklwood:
:+1: lgtm

PR #239, opened 2014-06-09
Feature/superadmin apis

albertshau:


docs/docs/source/rest/tenants.rst
```diff
@@ -0,0 +1,277 @@
...
+
+.. _tenants-delete:
+
+Delete a Tenant
+=================
+
+To delete a tenant, make a DELETE HTTP request to URI:
+::
+
+ /tenants/{name}
```
poornachandra: I think "name" should be "id" here.

docs/docs/source/rest/tenants.rst
```diff
@@ -0,0 +1,277 @@
...
+==================
+
+To update a tenant, make a PUT HTTP request to URI:
+::
+
+ /tenants/{id}
+
+Resource specified above respresents an individual tenant that is being updated.
+Currently, the update of a tenant resource requires the complete tenant object to be 
+returned back rather than individual fields.
```
poornachandra: The above line is a little confusing, I think we'll need to re-word it.

docs/docs/source/rest/tenants.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Example
+^^^^^^^^
+.. code-block:: bash
+
+ $ curl -X PUT
+        -H 'X-Loom-UserID:superadmin' 
+        -H 'X-Loom-Tenant:ID:loom'
+        -H 'X-Loom-ApiKey:<apikey>'
+        -d '{ "id":"f78dae92-a27b-4e3b-8c6a-cfc19f844259", "name":"my-company", "workers":20 }'
```
poornachandra, line 232: Can you update the name too in this request?

docs/docs/source/rest/tenants.rst
```diff
@@ -0,0 +1,277 @@
...
+
+Example
+^^^^^^^^
+.. code-block:: bash
+
+ $ curl -X GET 
+        -H 'X-Loom-UserID:superadmin' 
+        -H 'X-Loom-Tenant:ID:loom'
+        -H 'X-Loom-ApiKey:<apikey>'
+        http://<loom-server>:<superadmin-port>/<version>/tenants
```
poornachandra, line 276: It would be good to put in the response here.

server/src/main/java/com/continuuity/loom/http/LoomTenantHandler.java
```diff
@@ -0,0 +1,240 @@
...
+  @DELETE
+  @Path("/{tenant-id}")
+  public void deleteTenant(HttpRequest request, HttpResponder responder, @PathParam("tenant-id") String tenantId) {
+    if (!isSuperAdmin(request)) {
+      responder.sendStatus(HttpResponseStatus.FORBIDDEN);
+      return;
+    }
+
+    try {
+      store.deleteTenant(tenantId);
```
poornachandra, line 223: What if there are jobs running for the tenant? Is it safe to delete the tenant at any time?

poornachandra:

docs/docs/source/rest/tenants.rst
albertshau: good catch

docs/docs/source/rest/tenants.rst
albertshau: done

docs/docs/source/rest/tenants.rst
albertshau, line 232: yes, edited above to remove restriction on name changes.

docs/docs/source/rest/tenants.rst
albertshau, line 276: done

server/src/main/java/com/continuuity/loom/http/LoomTenantHandler.java
albertshau, line 223: The plan is to stop giving out additional tasks, but to allow deletion in the middle of jobs.  If an admin wishes to let all jobs finish, they need to enforce that for their users (maybe with some API to be added for disable cluster operations).

PR #245, opened 2014-06-13
Adding tenant to stores and APIs. To do this, introducing the

albertshau:
concept of an Account to keep track of all user information, and
views for the cluster and entity stores. Views are per account,
and only allow access to parts of the database that can be
accessed by that account.


server/src/main/java/com/continuuity/loom/account/Account.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.gson.annotations.Expose;
+
+/**
+ * An Account represents all information about a user, such as the user id and tenant id.
+ */
+public final class Account {
+  private final String userId;
+  private final String tenantId;
+  @Expose(serialize = false, deserialize = false)
+  private final boolean isSystem;
```
poornachandra, line 15: Can you add a comment on what isSystem means, and why it should not be serialized?

server/src/main/java/com/continuuity/loom/store/cluster/SQLClusterStoreService.java
```diff
@@ -0,0 +1,521 @@
...
+      DBQueryHelper.createDerbyTable("CREATE TABLE jobs ( cluster_id BIGINT, job_num BIGINT, status VARCHAR(32)," +
+                                       " create_time TIMESTAMP, job BLOB)",
+                                     dbConnectionPool);
+      DBQueryHelper.createDerbyTable("CREATE TABLE tasks ( cluster_id BIGINT, job_num BIGINT, task_num BIGINT," +
+                                       "submit_time TIMESTAMP, status_time TIMESTAMP, status VARCHAR(32), task BLOB )",
+                                     dbConnectionPool);
+      DBQueryHelper.createDerbyTable("CREATE TABLE nodes ( cluster_id BIGINT, id VARCHAR(64), node BLOB )",
+                                     dbConnectionPool);
+    }
+    systemView = new SQLSystemClusterStoreView(dbConnectionPool, Account.SYSTEM_ACCOUNT);
```
poornachandra: This can probably move to constructor.

server/src/main/java/com/continuuity/loom/store/cluster/SQLClusterStoreService.java
```diff
@@ -0,0 +1,521 @@
...
+      return new SQLUserClusterStoreView(dbConnectionPool, account);
+    }
+  }
+
+  @Override
+  public ClusterJob getClusterJob(JobId jobId) throws IOException {
+    try {
+      Connection conn = dbConnectionPool.getConnection();
+      try {
+        PreparedStatement statement = conn.prepareStatement("SELECT job FROM jobs WHERE job_num=? AND cluster_id=?");
```
poornachandra: Can you add a comment why job id is not filtered by tenant?

server/src/main/java/com/continuuity/loom/store/entity/SQLEntityStoreService.java
```diff
@@ -0,0 +1,67 @@
...
+import com.continuuity.loom.store.DBQueryHelper;
+import com.google.common.util.concurrent.AbstractIdleService;
+import com.google.inject.Inject;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
+ *
```
poornachandra: Empty javadoc

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
```diff
@@ -0,0 +1,222 @@
...
+  abstract PreparedStatement getSelectAllClustersStatement(Connection conn) throws SQLException;
+
+  abstract PreparedStatement getSelectClusterStatement(Connection conn, long id) throws SQLException;
+
+  abstract boolean allowedToWrite(Cluster cluster);
+
+  abstract PreparedStatement getSetClusterStatement(
+    Connection conn, long id, Cluster cluster, ByteArrayInputStream clusterBytes) throws SQLException;
+
+  abstract PreparedStatement getInsertClusterStatement(
```
poornachandra: getInsertClusterStatement can be implemented in the base class since it has the same implementation across all derived classes, right?

server/src/main/java/com/continuuity/loom/store/cluster/ClusterStoreService.java
```diff
@@ -0,0 +1,118 @@
...
+import com.continuuity.loom.scheduler.task.JobId;
+import com.continuuity.loom.scheduler.task.TaskId;
+import com.google.common.util.concurrent.Service;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ *
```
poornachandra: Empty javadoc

server/src/main/java/com/continuuity/loom/store/cluster/ClusterStoreService.java
```diff
@@ -0,0 +1,118 @@
...
+   */
+  ClusterStoreView getView(Account account);
+
+  /**
+   * Get a cluster job by its id.
+   * @param jobId Id of the cluster job to get.
+   * @return The cluster job with the given id, or null if none exists.
+   * @throws IOException if there was a problem getting the cluster job.
+   */
+  ClusterJob getClusterJob(JobId jobId) throws IOException;
```
poornachandra: Maybe move all methods that don't need a view to a different interface so that this interface is consistent with EntityStoreService? Also, the methods are used by different entities.

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
```diff
@@ -0,0 +1,222 @@
...
+        conn.close();
+      }
+    } catch (SQLException e) {
+      throw new IOException("Exception writing cluster " + cluster.getId(), e);
+    }
+  }
+
+  @Override
+  public void deleteCluster(String clusterId) throws IOException {
+    long clusterNum = Long.parseLong(clusterId);
```
poornachandra, line 161: writeCluster throws an exception when a user is not allowed to write to a cluster, but deleteCluster silently ignores it. We need to make the behavior uniform.

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
```diff
@@ -0,0 +1,222 @@
...
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Base abstract class for {@link ClusterStoreView} using a SQL database as the persistent store.
+ */
+public abstract class BaseSQLClusterStoreView implements ClusterStoreView {
+  private final DBConnectionPool dbConnectionPool;
+  private static final JsonSerde codec = new JsonSerde();
```
poornachandra: static final should be all CAPS.

poornachandra:
Mostly reviewed the new classes added, since others were just refactoring. Minor comments, please fix and merge.

server/src/main/java/com/continuuity/loom/account/Account.java
albertshau, line 15: done

server/src/main/java/com/continuuity/loom/store/cluster/SQLClusterStoreService.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/cluster/SQLClusterStoreService.java
albertshau: as mentioned further down, will move these methods into a separate interface that only the system will use.

server/src/main/java/com/continuuity/loom/store/entity/SQLEntityStoreService.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
albertshau: true, will move it

server/src/main/java/com/continuuity/loom/store/cluster/ClusterStoreService.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/cluster/ClusterStoreService.java
albertshau: agreed, these methods are never called by a user and would be cleaner to have them separate

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
albertshau, line 161: the difference is that write has a cluster object with an associated account when it is doing the writing.  Delete is just given an id, and I don't think we should distinguish between a user trying to delete a nonexistant cluster versus one they don't have access to.

server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
albertshau: done

PR #255, opened 2014-06-19
refactor guice modules into separate pieces so that we can

albertshau:
re-use them in smaller contexts.


poornachandra:
LGTM :+1:

PR #258, opened 2014-06-25
adding upgrade code for upgrading from 0.9.7 to 0.9.8.

albertshau:
Will keep a set of codecs for serializing/deserializing current
objects, and a separate set for converting between versions.
Switched to use guice injection to create the codec everywhere.


wolf31o2:
I think `loom-upgrade-tables-mysql.sql` should have versions in it. We're going to end up with a few of these, over time. Also, I think that we should do the same with the `loom-create-tables-mysql.sql` file. Either that, or we should merge them and make a single SQL that can create _or_ upgrade.

albertshau:
I can see why we'd want multiple upgrade files, but what would be the reason for multiple create files?  I don't think somebody would ever have to run more than one create file and if we want it for historical purposes they are always in git.

server/src/main/java/com/continuuity/loom/store/DBQueryHelper.java
```diff
@@ -26,19 +27,30 @@
...
       try {
         statement.executeUpdate(createString);
+        return true;
       } catch (SQLException e) {
-        // code for the table already exists in derby.
+        // code for the table already exists in derby. Since there is no create if not exists, we check the state
+        // and if the table already exists, we return false.  But if there was some other exception we'll throw that.
         if (!e.getSQLState().equals("X0Y32")) {
-          throw Throwables.propagate(e);
+          throw new SQLException(e);
```
poornachandra: You can throw the same object, right? Do you need to create a new wrapped exception?

albertshau: nope you're right.

server/src/main/java/com/continuuity/loom/http/handler/LoomAdminHandler.java
```diff
@@ -75,13 +74,13 @@
   public static final String PROVIDER_TYPES = "providertypes";
 
   private final EntityStoreService entityStoreService;
-  private final JsonSerde codec;
+  private final Gson gson;
 
   @Inject
-  private LoomAdminHandler(TenantStore tenantStore, EntityStoreService entityStoreService) {
+  private LoomAdminHandler(TenantStore tenantStore, EntityStoreService entityStoreService, JsonSerde codec) {
```
poornachandra: Should we replace all JsonSerde injection directly with Gson injection, since we use JsonSerde only to get a Gson object?

albertshau: aside from getting the gson, JsonSerde is only used to transform strings into bytes and vice versa, so we could just get rid of it and bind Gson in the codec module.  will do that.

server/src/main/java/com/continuuity/loom/tools/UpgradeTool.java
```diff
@@ -0,0 +1,93 @@
...
+      new ConfigurationModule(conf),
+      new StoreModule(),
+      new CodecModules().getUpgradeModule(NEW_TENANT_ID)
+    );
+    ClusterStoreService clusterStoreService = injector.getInstance(ClusterStoreService.class);
+    clusterStoreService.startAndWait();
+    DBConnectionPool dbConnectionPool = injector.getInstance(DBConnectionPool.class);
+    alterTables(dbConnectionPool);
+    migrateClusters(clusterStoreService);
+  }
```
poornachandra, line 41: We should stop the clusterStoreService at the end.

poornachandra:
One minor comment, you can merge after that :+1:

PR #260, opened 2014-06-25
use of the zookeeper namespace conf setting was accidentally

albertshau:
taken out during refactoring. Putting it back in.


wolf31o2:
:+1: LGTM

PR #261, opened 2014-06-26
concurrent zk queue test can take a really long time on really

albertshau:
slow machines.  Increasing the timeout and making it sleep for
less time.


poornachandra:
LGTM :+1:

PR #266, opened 2014-07-02
Feature/hadoop kerberos service

wolf31o2:
Add `hadoop-kerberos` service and add it to services which can use it. This requires #223 is resolved before it gets merged.


wolf31o2:
I am closing this. We will revisit it at a later time. For now, we just run the kerberos code in `hadoop_wrapper` all the time, which is a bit slow and requires external connections to the KDC.

PR #277, opened 2014-07-09
Feature/provisioner apis

albertshau:


server/src/main/java/com/continuuity/loom/common/queue/QueueType.java
```diff
@@ -24,7 +24,8 @@
   CLUSTER("/clustermanager/clustercreate"),
   JOB("/clustermanager/jobscheduler"),
   CALLBACK("/clustermanager/callback"),
-  PROVISIONER("/clustermanager/nodeprovision");
+  PROVISIONER("/clustermanager/nodeprovision"),
+  BALANCER("/clustermanager/balancer");
```
poornachandra, line 35: Now that we have so many queues, I think it would be good to have a brief description of each queue.

server/src/main/java/com/continuuity/loom/store/DBQueryHelper.java
```diff
@@ -143,6 +143,26 @@ public static void dropDerbyDB() {
   }
 
   /**
+   * Queries for a single number, returning the value of the number.
+   *
+   * @param statement PreparedStatement of the query, ready for execution.
+   * @return Result of the query.
```
poornachandra: @return Result of the query, or 0 if no results.

server/src/main/java/com/continuuity/loom/store/provisioner/ProvisionerStore.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * Persistent store for provisioner related information, such as capacity used.
+ */
+public interface ProvisionerStore extends Service {
+
+  Collection<Provisioner> getAllProvisioners() throws IOException;
```
poornachandra, line 20: Missing javadocs for interface methods.

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
```diff
@@ -0,0 +1,533 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implementation of the {@link ProvisionerStore} using a SQL database for persistent storage.  Maintains
+ * two tables, one for storing provisioner information and another for maintaining an index from tenants to
+ * provisioners.
+ */
+public class SQLProvisionerStore extends AbstractScheduledService implements ProvisionerStore {
```
poornachandra: Does SQLProvisionerStore need to be a AbstractScheduledService?

docs/docs/source/rest/provisioners.rst
```diff
@@ -0,0 +1,152 @@
...
+.. code-block:: bash
+
+ $ curl -X GET 
+        -H 'X-Loom-UserID:superadmin' 
+        -H 'X-Loom-Tenant:ID:loom'
+        -H 'X-Loom-ApiKey:<apikey>'
+        http://<loom-server>:<superadmin-port>/<version>/provisioners/p1
+ $ {
+       "id": "p1",
+       "host": "localhost",
```
poornachandra: localhost is probably not a good example.

server/src/main/java/com/continuuity/loom/http/handler/LoomProvisionerHandler.java
```diff
@@ -0,0 +1,164 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+
+/**
+ * Handles requests to take and finish tasks from provisioners.
```
poornachandra: The javadoc does not match the APIs

server/src/main/java/com/continuuity/loom/provisioner/Provisioner.java
```diff
@@ -0,0 +1,241 @@
...
+    Preconditions.checkArgument(host != null && !host.isEmpty(), "host must be specified.");
+    Preconditions.checkArgument(port != null, "port must be specified.");
+    Preconditions.checkArgument(capacityTotal != null && capacityTotal > 0,
+                                "capacityTotal must be specified and positive.");
+    this.id = id;
+    this.host = host;
+    this.port = port;
+    this.capacityTotal = capacityTotal;
+    this.usage = usage == null ? ImmutableMap.<String, Integer>of() : ImmutableMap.copyOf(usage);
+    this.assignments = assignments == null ? Maps.<String, Integer>newHashMap() : Maps.newHashMap(assignments);
```
poornachandra: It is probably better to use concurrent map here.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+  private TenantProvisionerService(ProvisionerStore provisionerStore,
+                                   TenantStore tenantStore,
+                                   ZKClient zkClient,
+                                   @Named(Constants.Queue.WORKER_BALANCE) TrackingQueue balanceQueue,
+                                   ProvisionerRequestService provisionerRequestService,
+                                   Configuration conf) {
+    this.provisionerStore = provisionerStore;
+    this.tenantStore = tenantStore;
+    this.provisionerRequestService = provisionerRequestService;
+    this.lock = new ZKInterProcessReentrantLock(zkClient, Constants.TENANT_NAMESPACE);
```
poornachandra, line 51: Can you add a brief description of why we use a single lock for all tenants?

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+  public void deleteTenant(String tenantId) throws IOException {
+    lock.acquire();
+    try {
+      Collection<Provisioner> tenantProvisioners = provisionerStore.getTenantProvisioners(tenantId);
+      provisionerStore.unassignTenantProvisioners(tenantId);
+
+      // send requests to provisioners to delete tenant
+      // if the db dies or something in between, the client making the request will get a 500.  No way to rollback,
+      // so client would have to resend the delete when things are ok again.
+      // TODO: what happens if we crash here? workaround would be to restart provisioners...
```
poornachandra: If the delete request goes on a queue then it will get retried when the server comes back up, right?

server/src/main/java/com/continuuity/loom/common/queue/QueueType.java
albertshau, line 35: done

server/src/main/java/com/continuuity/loom/store/DBQueryHelper.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/provisioner/ProvisionerStore.java
albertshau, line 20: done

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
albertshau: good catch, no it does not. Was part of an incomplete implementation that I didn't remove.

docs/docs/source/rest/provisioners.rst
albertshau: changed to "host1"

server/src/main/java/com/continuuity/loom/http/handler/LoomProvisionerHandler.java
albertshau: done

server/src/main/java/com/continuuity/loom/provisioner/Provisioner.java
albertshau: done

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau, line 51: done

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau: yeah though we'd need to put in the logic to retry queue elements and would have to make sure all the provisioners for the tenant get placed into a single queue element.  Will put that in a todo to be handled later.

albertshau: hm there are still race conditions since 2 tables are being updated plus the provisioner calls.  Another option would be to split this into 2 calls and only allow deleting a tenant if its workers are already set to 0.

poornachandra: Right.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+      lock.release();
+    }
+  }
+
+  public void writeProvisioner(Provisioner provisioner) throws IOException {
+    lock.acquire();
+    try {
+      // if a provisioner died earlier leaving us under capacity, and this provisioner adds capacity, we want to
+      // rebalance the workers after writing the provisioner.
+      boolean rebalanceWorkers = provisionerStore.getFreeCapacity() == 0;
```
poornachandra: Should we just re-balance every time?

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau: going to go with the 2nd option of forcing workers to be set to 0 before deleting.  Simplifies the logic and also makes it explicit to the caller that workers are stopped and no additional tasks will be kicked off when they delete the tenant.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau: agreed, this would be needed anyway if we have different balancing policies.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+        }
+      }
+    }
+  }
+
+  // TODO: abstract out to support different types of balancing policies
+  // Currently a greedy approach, just add to first available.
+  private void addWorkers(String tenantId, int numToAdd) throws CapacityException, IOException {
+    for (Provisioner provisioner : provisionerStore.getProvisionersWithFreeCapacity()) {
+      int numAdded = provisioner.tryAddTenantAssignments(tenantId, numToAdd);
```
poornachandra, line 295: You can add one more check - if numToAdd <= 0 then break;

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+        LOG.debug("Requesting provisioner {} to set workers to {} for tenant {} (adding {})",
+                  provisioner.getId(), provisioner.getAssignedWorkers(tenantId), tenantId, numAdded);
+        if (provisionerRequestService.putTenant(provisioner, tenantId)) {
+          numToAdd -= numAdded;
+        } else {
+          // request failed with retries. something is wrong with the provisioner, delete it and rebalance its workers
+          // TODO: what if this fails?
+          LOG.error("Could not write tenant {} to provisioner {}. The provisioner appears broken, deleting it and " +
+                     "rebalancing its tenant workers", tenantId, provisioner.getId());
+          deleteProvisioner(provisioner);
```
poornachandra, line 310: Can you add a comment on whether a rebalance will be triggered  during this call or after this call is complete.

server/src/main/java/com/continuuity/loom/provisioner/guice/ProvisionerModule.java
```diff
@@ -0,0 +1,17 @@
+package com.continuuity.loom.provisioner.guice;
+
+import com.continuuity.loom.provisioner.HttpProvisionerRequestService;
+import com.continuuity.loom.provisioner.ProvisionerRequestService;
+import com.google.inject.AbstractModule;
+import com.google.inject.Scopes;
+
+/**
+ *
```
poornachandra: Javadoc missing.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau, line 295: good point, done.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau, line 310: done

server/src/main/java/com/continuuity/loom/provisioner/guice/ProvisionerModule.java
albertshau: done

server/src/main/java/com/continuuity/loom/scheduler/Scheduler.java
```diff
@@ -154,5 +168,6 @@ private void unschedule() {
       }
     }
     scheduledFutures.clear();
+    tenantProvisionerService.stopAndWait();
```
poornachandra: Once a Guava service is stopped, it cannot be restarted again in schedule. Use a provider or something to get a new object the next time when it becomes leader.

poornachandra: Or else, make changes so that the service can keep running when server becomes follower.

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
```diff
@@ -0,0 +1,533 @@
...
+      } finally {
+        conn.close();
+      }
+    } catch (SQLException e) {
+      throw new IOException("Exception getting all provisioners", e);
+    }
+  }
+
+  @Override
+  public Collection<Provisioner> getIdleProvisioners(long idleTimestamp) throws IOException {
```
poornachandra: Idle does not convey the meaning of this method, right?

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -0,0 +1,289 @@
...
+
+      deleteProvisioner(provisioner);
+    } finally {
+      lock.release();
+    }
+  }
+
+  public void handleHeartbeat(String provisionerId, ProvisionerHeartbeat heartbeat)
+    throws IOException, MissingEntityException {
+    lock.acquire();
```
poornachandra: Is this lock required?

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
```diff
@@ -0,0 +1,533 @@
...
+          }
+          return results;
+        } finally {
+          rs.close();
+        }
+      } finally {
+        statement.close();
+      }
+    } finally {
+      statement.close();
```
poornachandra, line 423: I think there is an extra statement.close()

docs/docs/source/rest/tenants.rst
```diff
@@ -156,8 +156,8 @@ HTTP Responses
      - Description
    * - 200 (OK)
      - If delete was successful
-   * - 404 (NOT FOUND)
```
poornachandra, line 159: What happens when you don't find a tenant?

server/src/main/java/com/continuuity/loom/scheduler/Scheduler.java
albertshau: good catch.  Will move the provisioner timeout logic to a separate class so the service can always be running.

poornachandra:
Few more comments. You can merge after that is done :+1:

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
albertshau, line 423: yeah there is an extra try finally. will fix it.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
albertshau: good point, it is not required. will remove and add a comment.

docs/docs/source/rest/tenants.rst
albertshau, line 159: this was incorrect documentation to begin with.  Deleting a nonexistant entity returns a 200 so that it's idempotent.

server/src/main/java/com/continuuity/loom/store/provisioner/SQLProvisionerStore.java
albertshau: good point.  Will rename to something along the lines of 'unresponsive'

PR #285, opened 2014-07-16
adding a DBPut interface for standardizing how puts should be

albertshau:
done to the database, where a put is an update if the row already
exists or an insert if the row does not exist.

### DIFF
--- a/server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
+++ b/server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
@@ -20,13 +20,13 @@ import com.continuuity.loom.cluster.Node;
 import com.continuuity.loom.scheduler.task.ClusterJob;
 import com.continuuity.loom.store.DBConnectionPool;
 import com.continuuity.loom.store.DBHelper;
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/continuuity/loom/store/cluster/BaseSQLClusterStoreView.java
```diff
@@ -130,25 +125,12 @@ public void writeCluster(Cluster cluster) throws IllegalAccessException, IOExcep
...
-        }
-
-        // perform the update or insert
-        try {
-          writeStatement.executeUpdate();
-        } finally {
-          writeStatement.close();
-        }
+        DBPut clusterPut = new ClusterDBPut(clusterNum, cluster, clusterBytes);
+        dbQueryExecutor.executePut(conn, clusterPut);
```
poornachandra: I think we need to restructure this like we discussed to have uniform semantics for DBQueryExecutor. Since here the statement gets automatically closed in executePut, but in most other methods it does not. 

One more method where it gets automatically closed is getQuerySet.

Also - we document this behavior in DBQueryExecutor.

poornachandra:
Except for on comment - LGTM :+1:

PR #286, opened 2014-07-16
make automator and provider types readable across all tenants and

albertshau:
writeable by the superadmin only. This is because plugins are
currently constant across all tenants, but will change once that
is no longer the case.

### DIFF
--- a/server/src/main/java/com/continuuity/loom/http/handler/LoomAdminHandler.java
+++ b/server/src/main/java/com/continuuity/loom/http/handler/LoomAdminHandler.java
@@ -558,68 +558,6 @@ public class LoomAdminHandler extends LoomAuthHandler {
   }
 
   /**
-   * Delete a specific {@link ProviderType}. User must be admin or a 403 is returned.
-   *
-   * @param request The request to delete a provider type.
-   * @param responder Responder for sending the response.
-   * @param providertypeId Id of the provider type to delete.
-   */
-  @DELETE
-  @Path("/providertypes/{providertype-id}")
-  public void deleteProviderType(HttpRequest request, HttpResponder responder,
-                                 @PathParam("providertype-id") String providertypeId) {
-    Account account = getAndAuthenticateAccount(request, responder);
-    if (account == null) {
-      return;
-    }
-    if (!account.isSuperadmin()) {
-      responder.sendError(HttpResponseStatus.FORBIDDEN, "user unauthorized, must be superadmin.");
-      return;
-    }
-
-    try {
-      entityStoreService.getView(account).deleteProviderType(providertypeId);
-      responder.sendStatus(HttpResponseStatus.OK);
-    } catch (IOException e) {
-      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR,
-                          "Exception deleting provider type " + providertypeId);
-    } catch (IllegalAccessException e) {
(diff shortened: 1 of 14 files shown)


server/src/main/java/com/continuuity/loom/http/handler/LoomAdminHandler.java
```diff
@@ -572,8 +572,8 @@ public void deleteProviderType(HttpRequest request, HttpResponder responder,
     if (account == null) {
       return;
     }
-    if (!account.isAdmin()) {
-      responder.sendError(HttpResponseStatus.FORBIDDEN, "user unauthorized, must be admin.");
+    if (!account.isSuperadmin()) {
```
poornachandra: Now in this class - some methods need admin authorization, and some need super-admin authorization. I think it would be better to move super-admin methods into a separate handler.

server/src/main/java/com/continuuity/loom/store/entity/BaseSQLEntityStoreView.java
```diff
@@ -103,11 +104,24 @@
...
+    // immune to sql injection since everything is an enum or constant
+    StringBuilder queryStr = new StringBuilder();
+    queryStr.append("SELECT ");
+    queryStr.append(entityTypeId);
+    queryStr.append(" FROM ");
+    queryStr.append(entityTypeId);
+    queryStr.append("s WHERE name=? AND ");
+    // TODO: remove once types are defined through server instead of through provisioner
+    if (entityType == EntityType.AUTOMATOR_TYPE || entityType == EntityType.PROVIDER_TYPE) {
+      queryStr.append("( tenant_id=? OR tenant_id='");
```
poornachandra: Can you add a comment here?

server/src/main/java/com/continuuity/loom/store/entity/BaseSQLEntityStoreView.java
```diff
@@ -115,9 +129,21 @@ protected PreparedStatement getSelectStatement(Connection conn, EntityType entit
...
+    // immune to sql injection since everything is an enum or constant
+    StringBuilder queryStr = new StringBuilder();
+    queryStr.append("SELECT ");
+    queryStr.append(entityTypeId);
+    queryStr.append(" FROM ");
+    queryStr.append(entityTypeId);
+    queryStr.append("s WHERE tenant_id=? ");
+    // TODO: remove once types are defined through server instead of through provisioner
+    if (entityType == EntityType.AUTOMATOR_TYPE || entityType == EntityType.PROVIDER_TYPE) {
+      queryStr.append(" OR tenant_id='");
```
poornachandra: Comment here too.

server/src/test/java/com/continuuity/loom/http/LoomAdminHandlerTest.java
```diff
@@ -77,13 +78,13 @@ public void testClusterTemplates() throws Exception {
   @Test
   public void testProviderTypes() throws Exception {
     testNonPostRestAPIs("providertypes", Entities.ProviderTypeExample.JOYENT_JSON,
-                        Entities.ProviderTypeExample.RACKSPACE_JSON);
+                        Entities.ProviderTypeExample.RACKSPACE_JSON, SUPERADMIN_HEADERS);
```
poornachandra: Can you add a negative case.

poornachandra:
Few comments, you can merge after fixing them :+1:

PR #297, opened 2014-07-26
some test cleanup. Dont instantiate a scheduler to make sure it

albertshau:
does not start the schedulers and cause non deterministic
behavior. Also making zookeeper client retry for convenience when
debugging tests so that they dont fail when you set breakpoints
due to connection timeouts.


poornachandra:
LGTM :+1:

PR #298, opened 2014-07-26
fixing some more test cases that instantiate the scheduler.

albertshau:
Cant instantiate it and run a stop since the stop will be a no-op
if its not started.


poornachandra:
LGTM :+1:

PR #301, opened 2014-07-27
Feature/queue statistics

albertshau:


poornachandra:
Just one comment - right now the size() calls are expensive operations. It would be better to make it return from a cached value so that every get size is constant time operation.

PR #302, opened 2014-07-27
Enforcing max cluster and max node settings within a tenant.

albertshau:
Trying to create a cluster that violates the max will cause the
request to fail with a 409. Similarly, trying to lower the max
below whats already in use will cause the request to fail with a
409.


docs/docs/source/rest/tenants.rst
```diff
@@ -181,7 +182,8 @@ To update a tenant, make a PUT HTTP request to URI:
 
 The resource specified above respresents an individual tenant that is being updated.
 Currently, the update of a tenant resource requires the complete tenant object to in
-the request body. 
+the request body. Trying to lower the max clusters or max nodes belong the number
```
poornachandra: belong -> below?

poornachandra: But what if as an administrator I want to reduce the tenant's quota below what the tenant is using right now? If you allow me to reduce the quota even if it causes violations now, later when the tenant deletes some cluster he won't be able to replace them. This will automatically cause the quota to be respected at that point. 

If you don't allow the admin to reduce the quota then he'll never be able to reduce a tenant's quota, right?

albertshau: The reasoning behind failing the request in this case is that setting a maximum implies intent for that to be a hard maximum.  If you are able to lower the max below what is in use now, what was the point of setting the max that low to begin with?

server/src/main/java/com/continuuity/loom/scheduler/task/ClusterService.java
```diff
@@ -101,34 +107,46 @@ public ClusterService(ClusterStoreService clusterStoreService,
...
-    LOG.debug("adding create cluster element to solverQueue");
-    SolverRequest solverRequest = new SolverRequest(SolverRequest.Type.CREATE_CLUSTER,
-                                                    gson.toJson(clusterCreateRequest));
-    solverQueues.add(account.getTenantId(), new Element(cluster.getId(), gson.toJson(solverRequest)));
-
-    loomStats.getClusterStats().incrementStat(ClusterAction.SOLVE_LAYOUT);
-    return cluster.getId();
+    throws IOException, IllegalAccessException, QuotaException {
+    ZKInterProcessReentrantLock lock = new ZKInterProcessReentrantLock(zkClient, "create");
+    lock.acquire();
```
poornachandra, line 118: Add a comment on why the lock is required.

poornachandra:

PR #304, opened 2014-07-28
Feature/bootstrap tenant

albertshau:


docs/docs/source/rest/rpc.rst
```diff
@@ -24,8 +24,67 @@ RPC
...
+.. _rpc-bootstrap:
+
+Bootstrap Tenant
+================
+Bootstrapping a tenant will copy all providers, hardware types, image types, services, 
+cluster templates, and plugin resources from the superadmin to the tenant making the request.
+The user making the request must be the tenant admin. This is a copy, meaning if the 
+superadmin updates the entities, the changes will not be reflected in the tenant.
+
+To bootstrap a tenant, making a POST HTTP request to URI:
```
dereklwood, line 38: no URI... incomplete?

docs/docs/source/rest/rpc.rst
```diff
@@ -24,8 +24,67 @@ RPC
...
+::
+
+ /bootstrap
+
+The body is a JSON Object containing one optional key ``overwrite`` which indicates whether or 
+not data from the superadmin should overwrite existing data. By default it is false, meaning an entity
+from the superadmin will not skipped if it already exists in the tenant account, even if the entities
```
dereklwood: ...will _be_ skipped... (doublecheck)

albertshau: yep. fingers are auto-completing poorly.

docs/docs/source/rest/rpc.rst
albertshau, line 38: the path is just /v1/bootstrap which is below

docs/docs/source/rest/rpc.rst
```diff
@@ -24,8 +24,67 @@ RPC
...
+================
+Bootstrapping a tenant will copy all providers, hardware types, image types, services, 
+cluster templates, and plugin resources from the superadmin to the tenant making the request.
+The user making the request must be the tenant admin. This is a copy, meaning if the 
+superadmin updates the entities, the changes will not be reflected in the tenant.
+
+To bootstrap a tenant, making a POST HTTP request to URI:
+::
+
+ /bootstrap
```
poornachandra, line 41: Can /bootstrap be called on a tenant multiple times? If so we should change the name to something else, right?

server/src/main/java/com/continuuity/loom/http/handler/LoomRPCHandler.java
```diff
@@ -54,18 +61,98 @@
...
+   * @param request Request to bootstrap the tenant
+   * @param responder Responder for responding to the request
+   * @throws Exception
+   */
+  @POST
+  @Path("/bootstrap")
+  public void bootstrapTenant(HttpRequest request, HttpResponder responder) throws Exception {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
```
poornachandra, line 103: Is is possible to get a request with empty account?

poornachandra:
Minor comments, LGTM :+1:

server/src/main/java/com/continuuity/loom/http/handler/LoomRPCHandler.java
albertshau, line 103: It's possible headers are missing, which is checked in getAndAuthenticateAccount.  The null check is really a check to see if that function found something wrong.

docs/docs/source/rest/rpc.rst
albertshau, line 41: true.  will change this to only allow bootstrap if everything is completely empty.  If we want to support tenants using superadmin entities and having those entities update as the superadmin updates them, it should be something separate.  Maybe adding links or inheritance or something along those lines.

PR #306, opened 2014-07-29
Feature/use tenant name externally

albertshau:


server/src/main/java/com/continuuity/loom/provisioner/Provisioner.java
```diff
@@ -47,6 +49,23 @@ public Provisioner(String id, String host, Integer port, Integer capacityTotal,
     Preconditions.checkArgument(this.capacityFree >= 0, "assignments cannot exceed total capacity");
   }
 
+  public static Provisioner swapIdsForNames(Provisioner provisioner, LoadingCache<String, String> idNameMap)
```
poornachandra: javadoc missing for public method.

poornachandra: Since it creates a new Provisioner, we should probably rename it to reflect that.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -51,70 +59,101 @@ private TenantProvisionerService(ProvisionerStore provisionerStore,
     this.lock = new ZKInterProcessReentrantLock(zkClient, Constants.TENANT_NAMESPACE);
     this.provisionerTimeoutSecs = conf.getLong(Constants.PROVISIONER_TIMEOUT_SECS);
     this.balanceQueue = balanceQueue;
+    tenantIdNameMap = CacheBuilder.newBuilder().build(new CacheLoader<String, String>() {
```
poornachandra: As you suggested, it would be better to push the cache to the Store layer.

poornachandra:

server/src/main/java/com/continuuity/loom/provisioner/Provisioner.java
albertshau: done

PR #311, opened 2014-07-30
Simple cluster template for a base system-only

wolf31o2:


config/defaults/clustertemplates/base.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "base",
+  "description": "Base system",
```
albertshau: whats the intent of this template? To test that a provider is working? Can we have a more descriptive description

albertshau:
minor comment, lgtm

config/defaults/clustertemplates/base.json
wolf31o2: I'll update the description... This is pretty much the closest to a no-op cluster that current Loom can handle.

PR #312, opened 2014-07-30
Feature/more hardwaretypes

wolf31o2:
Add hardwaretypes for xlarge/xxlarge for larger clusters


albertshau:
lgtm. We should eventually rethink the hardware type names.

wolf31o2:
I agree. The problem is that there's not really any kind of standard between providers on how they name their flavors... AWS/OpenStack have (somewhat) parity... as does Rackspace for standard types. Joyent does their own thing, completely.

PR #317, opened 2014-07-30
Feature/php module support

wolf31o2:
Takes a list of PHP PEAR modules to install via attributes... Added to lamp clustertemplate... Fixes #316


albertshau:
lgtm

PR #319, opened 2014-07-31
Feature/provisioner configfile

dereklwood:
- [x] provisioner reads properties from a default config file, and optionally, an additional override file given with the `-c <file>` option
- [x] additional configs such as log rotation settings
- [x] removed all command line arguments except `-c <configfile>` and `--register`
- [ ] we may consider changing config files to .rb... xml was just easy to implement first.
- [x] minor unrelated change: abort startup if unhandled exceptions in heartbeat thread


provisioner/master/conf/provisioner-default.xml
```diff
@@ -0,0 +1,70 @@
...
+    <value>/tmp/loom/data</value>
+    <description>Provisioner storage directory for plugin data resources</description>
+  </property>
+  <property>
+    <name>provisioner.work.dir</name>
+    <value>/tmp/loom/work</value>
+    <description>Provisioner working directory</description>
+  </property>
+  <property>
+    <name>provisioner.default.capacity</name>
```
albertshau: can we remove default from the setting name

provisioner/master/lib/provisioner/cli.rb
```diff
@@ -25,40 +25,13 @@ def self.read(arguments = ARGV)
       options = {}
       OptionParser.new do |opts|
         opts.banner = 'Usage: '
-        opts.on('-u', '--uri URI', 'Loom web server URI') do |u|
-          options[:uri] = u
+        opts.on('-C', '--config-file FILE', 'Site-specific config file to use') do |c|
```
albertshau: would prefer lowercase for the option

provisioner/master/lib/provisioner/config.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def initialize(options)
+      @properties = {} # name => value
+      @descriptions = {} # name => description
+      @default_file = "#{File.dirname(__FILE__)}/../../conf/provisioner-default.xml"
+      @site_file = options[:configfile]
+    end
+
+    def get(val)
+      val = 'provisioner.' + val.downcase unless val.downcase =~ /^provisioner\./i
```
albertshau: I think its better to just use the given input as the name without prepending 'provisioner.'.  It makes the behavior more consistent, otherwise we might end up with places in the code where somebody looks up 'log.directory' and other places where somebody looks up 'provisioner.log.directory'

provisioner/master/lib/provisioner/config.rb
```diff
@@ -0,0 +1,98 @@
...
+                p_value = child.text
+              elsif child.name == 'description'
+                p_description = child.text
+              end
+            end
+
+            # ignore anything without a name and value
+            next if p_name.nil? || p_value.nil?
+            # only consider 'provisioner.*' entries
+            next unless p_name.downcase =~ /^provisioner\./i
```
albertshau: I don't think we need to enforce this. For example if we ever changed the provisioner to read directly from the queue instead of calling the server, we might have config settings that don't fit into this pattern.

provisioner/master/lib/provisioner/provisioner.rb
```diff
@@ -50,15 +52,24 @@ def initialize(options)
...
+      config = Config.new(options)
+      config.load
+
       # initialize logging
-      Logging.configure(options[:log_directory] ? "#{options[:log_directory]}/provisioner.log" : nil)
-      Logging.level = options[:log_level]
-      Logging.log.info "Loom api starting up"
+      Logging.configure(config.get('log.directory') ? "#{config.get('log.directory')}/provisioner.log" : nil)
+      Logging.level = config.get('log.level')
+      Logging.shift_age = config.get('log.rotation.shift.age')
```
albertshau: would be better to move the strings to some constants file and use variables everywhere, so that if we ever change the name of a setting we just have to change it in one place.

provisioner/master/spec/provisioner-default.xml
```diff
@@ -0,0 +1,98 @@
+<configuration>
```
albertshau, line 1: how come there are 2 default files?

albertshau:
couple comments, looks good to merge after that.  Merging develop should fix the travis failure too.

provisioner/master/spec/provisioner-default.xml
dereklwood, line 1: this one is in the test directory, just for testing

provisioner/master/lib/provisioner/tenantspec.rb
```diff
@@ -25,7 +25,7 @@ class TenantSpec
     def initialize(id, workers, resources = nil, plugins = nil)
       @id = id
       @workers = workers
-      # @resources = resources ||= Hash.new { |h, k| h[k] = Hash.new(&h.default_proc) }
+      @resources = resources ||= Hash.new { |h, k| h[k] = Hash.new(&h.default_proc) }
```
wolf31o2, line 28: Is this being used somewhere?

dereklwood, line 28: not yet, but changing it here would just introduce a merge conflict in my next pending branch :)

PR #320, opened 2014-08-01
Feature/allow non root bootstrap

wolf31o2:
This simply determines if a user is not root, and uses sudo before privileged commands, if necessary. This is required for #314 and covers the plugins.


albertshau:
lgtm

PR #326, opened 2014-08-01
make sure the resource type exists before actually uploading it

albertshau:
or trying to perform operations on it, returning a 404 if it
does not exist.


poornachandra:
LGTM :+1:

PR #328, opened 2014-08-01
temporary workaround to include a ssh-user field in the image type

albertshau:
This is the server side of #314


wolf31o2:
:+1: LGTM

PR #329, opened 2014-08-01
internal provisioner api will be trusted when security is added,

albertshau:
should not be authenticating with an api key and provisioners
should be able to get any tenant resource they needed. Thus,
moving tenant id into the path.


poornachandra:
LGTM :+1:

PR #331, opened 2014-08-01
fix mysql table schemas. Changing foreign key on tasks table so

albertshau:
it doesn't break, removing indexes on status since status is
low cardinality, and reducing length of columns if possible.

### DIFF
--- a/config/sql/loom-create-tables-mysql.sql
+++ b/config/sql/loom-create-tables-mysql.sql
@@ -117,16 +117,16 @@ CREATE TABLE IF NOT EXISTS tenants (
     name VARCHAR(255),
     workers INT,
     deleted BOOLEAN,
-    create_time TIMESTAMP,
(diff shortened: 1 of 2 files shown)


config/sql/loom-create-tables-mysql.sql
```diff
@@ -127,8 +136,8 @@ CREATE TABLE IF NOT EXISTS provisioners (
 ) ENGINE = InnoDB;
 
 CREATE TABLE IF NOT EXISTS provisionerWorkers (
-    provisioner_id VARCHAR(255),
-    tenant_id VARCHAR(255),
+    provisioner_id VARCHAR(64),
```
wolf31o2, line 139: Making sure @dereklwood knows that this is only a max of 64 characters...

wolf31o2:
:+1: LGTM

PR #332, opened 2014-08-02
switch to use setBytes whenever we are writing a blob since some

albertshau:
jdbc implementations have trouble with input streams (looking at
you mysql).


poornachandra:
LGTM :+1:

PR #334, opened 2014-08-04
Feature/hadoop updates zk security

wolf31o2:
Import-only PR to update hadoop/hadoop_wrapper to support zookeeper server kerberos authentication


albertshau:
lgtm

PR #335, opened 2014-08-04
change superadmin tenant to superadmin instead of loom. It is a

albertshau:
more meaningful name and we shouldn't be putting the product name
in so many places.


wolf31o2:
How about load-defaults.sh and friends?

albertshau:
good point, fixed.  searched for 'loom' and "loom" and think I got everywhere.

wolf31o2:
:+1: LGTM... worst case is we missed a few places

PR #336, opened 2014-08-05
ability to disable multi-tenancy, in which case the tenant header

albertshau:
is ignored and the superadmin tenant is used everywhere a tenant
would normally be used.

### DIFF
--- a/docs/docs/source/guide/admin/server-config.rst
+++ b/docs/docs/source/guide/admin/server-config.rst
@@ -244,6 +244,9 @@ A full list of available configuration settings and their default values are giv
    * - server.callback.http.max.connections
      - 100
      - max number of concurrent http connections for callbacks. If the max is reached, the next callback to try and send a request blocks until an open connection frees up.
+   * - server.multitenancy.enabled
+     - true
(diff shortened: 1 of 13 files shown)


wolf31o2:
Should we set `server.multitenancy.enabled` to `false` by default for the singlenode/standalone?

albertshau:
My first inclination is yes since the idea with standalone is to try and get up and running as simply as possible.

wolf31o2:
I agree. Want to change that? Otherwise, LGTM... might want to get a Java dev to peek at it, though.

albertshau:
I'm starting to wonder if this is more confusing than its worth.  Leaning towards not having this.

wolf31o2:
Well, now that superadmin has cluster operation privileges, there's no reason someone couldn't have a single tenant and just use that to achieve single tenant status.

albertshau:
yup. The inconsistencies this would introduce (how does one change # workers, max clusters, max nodes, and other tenant settings in a non-multi-tenant world? what about the UI?) outweigh the small convenience it brings. Abandoning this.

PR #340, opened 2014-08-05
Remove ALL RIGHTS RESERVED from footer. Fixes #123

wolf31o2:
We use the Apache License, v2.0... I have removed the 'ALL RIGHTS RESERVED' link, but am unsure if we should link to the [Apache License](http://www.apache.org/licenses/LICENSE-2.0.html), instead.


albertshau:
lgtm

PR #342, opened 2014-08-06
fix case where the zkmap has not yet loaded an additional key

albertshau:
and so we cant assume the value for the key exists.


poornachandra:
LGTM :+1:

PR #345, opened 2014-08-07
Feature/node refactor

albertshau:


server/src/main/java/com/continuuity/loom/cluster/Node.java
```diff
@@ -67,34 +49,18 @@
...
-        }
-      }
-      serviceNames.add(new JsonPrimitive(service.getName()));
-    }
-    this.properties.add(Properties.AUTOMATORS.name().toLowerCase(), automators);
-    this.properties.add(Properties.SERVICES.name().toLowerCase(), serviceNames);
-
-    this.actions = Lists.newArrayList();
+    this.properties = properties;
+    this.actions = actions;
```
poornachandra: Don't we need null check for properties and actions.

server/src/main/java/com/continuuity/loom/cluster/NodeProperties.java
```diff
@@ -0,0 +1,315 @@
...
+   * @param nodenum Number of the node in a cluster. Each node in a cluster has a different number.
+   * @param hardwaretype Name of the hardware type of the node.
+   * @param imagetype Name of the image type of the node.
+   * @param flavor Flavor of the node.
+   * @param image Image of the node.
+   * @param sshUser User the provisioner will use to ssh in to the node.
+   * @param services Set of {@link Service services} on the node.
+   * @return Node properties created from the given input.
+   */
+  public static NodeProperties from(String hostname, String ipaddress, int nodenum, String hardwaretype,
```
poornachandra: Maybe this should be part of the builder?

poornachandra:

PR #348, opened 2014-08-07
Mount data disk on Rackspace

wolf31o2:
This checks for the presence of `/dev/xvde1` which is where Rackspace mounts larger data disks on their performance line of hardware types. This takes that disk and mounts it into `/data` so that it is usable.

### DIFF
--- a/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/rackspace.rb
+++ b/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/rackspace.rb
@@ -102,7 +102,7 @@ class FogProviderRackspace < Provider
       # Validate connectivity and Mount data disk
       Net::SSH.start(@result['result']['ipaddress'], @task['config']['ssh-auth']['user'], @credentials) do |ssh|
         ssh_exec!(ssh, 'ping -c1 www.opscode.com', 'Validating external connectivity and DNS resolution via ping')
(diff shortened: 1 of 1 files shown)


provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/rackspace.rb
```diff
@@ -95,13 +95,14 @@ def confirm(inputmap)
...
       @result['result']['ipaddress'] = bootstrap_ip
-      # Additional checks
+      # do we need sudo bash?
+      sudo = 'sudo' unless @task['config']['ssh-auth']['user'] == 'root'
       set_credentials(@task['config']['ssh-auth'])
-      # Validate connectivity
+      # Validate connectivity and Mount data disk
       Net::SSH.start(@result['result']['ipaddress'], @task['config']['ssh-auth']['user'], @credentials) do |ssh|
         ssh_exec!(ssh, 'ping -c1 www.opscode.com', 'Validating external connectivity and DNS resolution via ping')
+        ssh_exec!(ssh, "test -e /dev/xvde1 && (#{sudo} /sbin/mkfs.ext4 /dev/xvde1 && #{sudo} mkdir -p /data && #{sudo} mount /dev/xvde
```
dereklwood: think we need an `|| true` or similar at the end.  if the box doesn't have /dev/xvde1, ssh_exec will return 1 and fail the task

wolf31o2: It was worse than that! When I copied/pasted this, I missed part of the line... Oops!

dereklwood:
:+1:

PR #350, opened 2014-08-08
Check number of arguments to ssh_exec and do the right thing

wolf31o2:
This allows the `fog_provisioner` to be used on older versions of Loom, such as the 0.9.7 standalone...


dereklwood:
referring to previous PR #229 

CommandExecutionError method signature also changed.  So what will happen now in failure scenario is that it will fail when trying to create a new exception to throw.  On the UI it will look like "unexpected error, wrong number of arguments..." ie a code issue

PR #351, opened 2014-08-08
Feature/issue 313

wolf31o2:
This fixes #313 for AWS/Joyent... it seems that these providers have some default DROP firewall, so initial TCP connections to the host hang until TCP's default timeout.


albertshau:
seems like there should be a better way, but lgtm for now

wolf31o2:
Yeah... the problem is that TCPSocket.new doesn't allow setting a timeout. I have tried a couple other options, and this works, for now. I will need to end up rewriting all of this using Socket.new, instead. However, I would prefer not change that this late in the release cycle, when a small sleep can overcome the issue.

PR #352, opened 2014-08-08
Feature/provisioner resourcemgr

dereklwood:
first part of data syncing.  this PR adds capability to master process to act upon the resources block in the tenant api json.  Upon any changes or startup, master will shutdown any workers, fetch resources from server via api, sync locally, activate, and resume workers.  At this point workers are still pointing to their own data, next PR will change that.
- [x] resource thread in main provisioner to do the asynchronous coordination
- [x] some changes to the handling of uncaught exceptions in threads, should gracefully shutdown
- [x] resourcespec as part of tenantspec, resourcemanager to do all the actual operations
- [x] updated the current plugin definitions with changes made in config/defaults


provisioner/master/lib/provisioner/api.rb
```diff
@@ -16,40 +16,39 @@
...
         begin
-          log.info "heartbeat called"
           settings.provisioner.heartbeat.to_json
         rescue
           halt 500
         end
       end
 
-      post "/v1/tenants" do
+      post '/v1/tenants' do
```
albertshau: it's better to use a constant for the version so we just have to change it in one place when it gets updated (which is soon).

provisioner/master/lib/provisioner/provisioner.rb
```diff
@@ -185,8 +204,23 @@ def spawn_heartbeat_thread
...
-      # abort on any uncaught exception during registration, etc
-      @heartbeat_thread.abort_on_exception=true
+    end
+
+    def spawn_resource_thread
+      @resource_thread = Thread.new {
+        log.info "starting resource thread"
+        loop {
+          @tenantmanagers.each do |id, tmgr|
+            if tmgr.resource_sync_needed? && tmgr.num_workers == 0
```
albertshau, line 212: we need to handle the case where another sync is called while we are syncing.  We could make this a while resource_sync_needed and then resume after the loop.

provisioner/master/lib/provisioner/resourcespec.rb
```diff
@@ -0,0 +1,65 @@
...
+    #   "active"=>[{"name"=>"reactor", "version"=>"2"}, {"name"=>"hadoop", "version"=>"52"}]}}}}
+    def initialize(resource_jsonobj)
+      # resources is a hash of '/'-delimited resource paths and their versions
+      #   ex: "automatortype/chef-solo/cookbooks/hadoop" => 2
+      @resources = {}
+      # resource_formats is a hash of '/'-delimited resource paths and their formats
+      #   ex: "automatortype/chef-solo/cookbooks/hadoop" => "archive"
+      @resource_formats = {}
+      unless resource_jsonobj.nil?
+        resource_jsonobj.each do |type, h_type|
```
albertshau, line 36: a comment at each nested loop would make it easier to understand the structure of the json

provisioner/master/lib/provisioner/provisioner.rb
dereklwood, line 212: albert, addressed stacked sync calls in https://github.com/continuuity/loom/commit/78e7017af2bf1549d2978f4bd9a9023bd31519c3 if you can take a look

albertshau:
lgtm

PR #347, opened 2014-08-07
Add new ipaddresses hash for issue #343

wolf31o2:
- [x] Joyent, OpenStack, Rackspace
- [x] AWS (done in another PR)
- [x] Provisioner Automator
- [x] Server-side support
- [x] UI changes to new addresses
- [x] Cluster templates
- [x] Docs updated

Fixes #343

PR #358, opened 2014-08-11
provisioner generated fields moved out of node properties and into

albertshau:
its own section. As a result, the user and password shown in the
ui for the rackspace provider needs to change.


wolf31o2:
This will need to be refactored when #347 is merged.

PR #359, opened 2014-08-11
Feature/standalone startup race

dereklwood:
fixes a race condition in standalone startup:

```
Starting Loom Provisioner ...
Requesting 5 workers for default tenant...
Not enough capacity to update tenant.
```

this was because the status check used by wait_for_provisioner returns OK as soon as API is up, which could be before the heartbeat thread registers with the server.
- [x] make the /status call meaningful, returns status as updated during startup procedure
- [x] heartbeat thread now tracks once there is a successful registration, sets status, and will continuallly retry upon startup.  this should allow provisioner be started up before server.


albertshau:
lgtm

PR #360, opened 2014-08-11
Feature/add node api

seawatts:


README.md
```diff
@@ -18,6 +18,47 @@ To learn more about Continuuity Loom, here are the resources available:
       * [Documentation 0.9.6](http://www.continuuity.com/docs/loom/0.9.6/en/index.html)
       * [Documentation 0.9.5](http://www.continuuity.com/docs/loom/0.9.5/en/index.html)
 
+## Building Source
```
albertshau, line 21: nice, thanks for adding this

README.md
```diff
@@ -18,6 +18,47 @@ To learn more about Continuuity Loom, here are the resources available:
...
+
+**Backend Java Server**
+_Prerequisites_
```
albertshau, line 25: when you view the rendered file, is prerequisites showing up on the same line as "backend java server" for you?

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+import java.io.Reader;
+import java.util.Set;
+
+/**
+ * Handler for performing node operations.
+ */
+@Path("/v1/loom/nodes")
+public class LoomNodeHandler extends LoomAuthHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(LoomNodeHandler.class);
+  //  private final NodeService clusterService;
```
albertshau: can remove this if its not needed

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+  private final NodeStoreService nodeStoreService;
+  private final NodeStore nodeStore;
+  private final Gson gson;
+
+  /**
+   * Initializes a new instance of a LoomNodeHandler.
+   */
+  @Inject
+  public LoomNodeHandler(TenantStore tenantStore, NodeService nodeService, NodeStoreService nodeStoreService,
```
albertshau: It's better if the constructor is private to prevent people from trying to new it.

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+                         Gson gson) {
+    super(tenantStore);
+    this.nodeStoreService = nodeStoreService;
+    this.nodeStore = this.nodeStoreService.getSystemView();
+    this.gson = gson;
+  }
+
+  /**
+   * Get all noes visible to the user.
```
albertshau: 'nodes' typo

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+
+    Set<Node> nodes = null;
+    try {
+      nodes = nodeStoreService.getView(account).getAllNodes();
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception getting nodes.");
```
albertshau, line 87: we can return here to prevent null pointers further on

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    JsonArray jsonArray = new JsonArray();
+    for (Node node : nodes) {
+      JsonObject obj = new JsonObject();
+      obj.addProperty("id", node.getId());
+      obj.addProperty("clusterId", node.getClusterId());
+      //      obj.add("properties", node.getProperties());
```
albertshau: we don't want properties here?

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    responder.sendJson(HttpResponseStatus.OK, jsonArray);
+  }
+
+  @POST
+  public void createNodes(HttpRequest request, HttpResponder responder) {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
```
albertshau, line 124: should close the reader

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    Set<Node> nodes = gson.fromJson(reader, new TypeToken<Set<Node>>(){}.getType());
+    try {
+      nodeStoreService.getView(account).writeNodes(nodes);
+    } catch (IllegalAccessException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception creating nodes.");
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception creating nodes.");
+    }
+
+    responder.sendStatus(HttpResponseStatus.CREATED);
```
albertshau: should move this into the try block. Otherwise, if there's an error writing we'll write to the responder once for the error and then again here (which will result in an exception)

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+    Set<Node> nodes = gson.fromJson(reader, new TypeToken<Set<Node>>(){}.getType());
+    try {
+      nodeStoreService.getView(account).writeNodes(nodes);
+    } catch (IllegalAccessException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception creating nodes.");
```
albertshau: would be more accurate to return a forbidden status since this exception indicates the account cant access the resource

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+    try {
+      nodeStoreService.getView(account).deleteNode(nodeId);
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception deleting node.");
+    }
+
+    responder.sendStatus(HttpResponseStatus.NO_CONTENT);
+  }
+}
```
albertshau, line 208: should there be an endpoint to get a specific node?

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    } catch (IllegalAccessException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception creating nodes.");
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception creating nodes.");
+    }
+
+    responder.sendStatus(HttpResponseStatus.CREATED);
+  }
+
+  @PUT
```
albertshau, line 152: the PUT should contain the node id in the path, since we're writing a single nodes and not the collection of nodes.

server/src/test/java/com/continuuity/loom/http/LoomNodeHandlerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    Assert.assertEquals(0, nodes.size());
+  }
+
+  @Test
+  public void testUpdateNodeAsUser() throws Exception {
+    Node node = postNodes(2, USER1_HEADERS).get(0);
+    String nodeJsonString = getNodeAsJsonString(node);
+    HttpResponse response = doPut("/v1/loom/nodes", nodeJsonString, USER1_HEADERS);
+    assertResponseStatus(response, HttpResponseStatus.NO_CONTENT);
+    // test to see that it has been updated
```
albertshau: can you add this part

README.md
seawatts, line 21: No problem. Not fully finished though i guess didnt put in the Running source part or Running in IntelliJ

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
seawatts: Oops forgot to add those

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,199 @@
...
+  @Path("/{node-id}")
+  public void updateNode(HttpRequest request, HttpResponder responder, @PathParam("node-id") String nodeId) {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+    Node node = gson.fromJson(reader, Node.class);
```
albertshau: should check if the node we're given in the body has the same id as whats in the path and 400 if they're different.

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,199 @@
...
+    Set<Node> nodes;
+    try {
+      nodes = nodeStoreService.getView(account).getAllNodes();
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception getting nodes.");
+      return;
+    }
+
+    JsonArray jsonArray = new JsonArray();
+    for (Node node : nodes) {
```
albertshau: actually, I think it will work if you send the set directly in sendJson

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
albertshau: should 400 if there is any other exception reading the body too.

albertshau:
couple more small comments, otherwise lgtm

seawatts:
Alright, fixed those last couple comments

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,193 @@
...
+  public void updateNode(HttpRequest request, HttpResponder responder, @PathParam("node-id") String nodeId) {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+
+    try {
+      Node node = gson.fromJson(reader, Node.class);
+      if (node.getId().equals(nodeId)) {
```
albertshau, line 168: Since nodes can be input by users, we'll need to check that the id is not null.  I think it makes sense to put the check in the node constructor since id shouldn't ever be missing.

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  @POST
+  public void createNodes(HttpRequest request, HttpResponder responder) {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+    Reader reader = new InputStreamReader(new ChannelBufferInputStream(request.getContent()), Charsets.UTF_8);
+    Set<Node> nodes = gson.fromJson(reader, new TypeToken<Set<Node>>() {}.getType());
```
albertshau: I think it's simpler to only accept one node at a time instead of multiple.  Otherwise we could have situations where some node in the middle of the set is invalid (maybe the id already exists or something like that).  Or somebody posts in a humongous set and the server OOMs.

albertshau: here we'll also have to check that the body is valid json and 400 if not.

albertshau:
couple more comments, then lgtm.

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -0,0 +1,193 @@
...
+        responder.sendError(HttpResponseStatus.BAD_REQUEST, "Node ID in body does not match Node ID in URI path");
+      }
+    } catch (IllegalAccessException e) {
+      responder.sendError(HttpResponseStatus.FORBIDDEN, "Exception updating node.");
+    } catch (IOException e) {
+      responder.sendError(HttpResponseStatus.INTERNAL_SERVER_ERROR, "Exception updating node.");
+    } catch (JsonIOException e){
+      responder.sendError(HttpResponseStatus.BAD_REQUEST, "Exception reading node from body.");
+    } catch (JsonSyntaxException e){
+        responder.sendError(HttpResponseStatus.BAD_REQUEST, "Exception reading node from body.");
```
albertshau: indentation

seawatts:
Alright fixed comments

albertshau:
:+1:

PR #368, opened 2014-08-13
Feature/provisioner workerdata

dereklwood:
this is the simplest, most direct way to point the workers at the synced data directory.  I'd like to revisit a lot of the worker, but this will work in the short-term.
- [x] --work-dir passed to worker as an argument
- [x] plugins are initialized with an `@env` instance var, which can be used to retrieve all configuration options in the worker.  this is one place we could probably figure out a better interface for the plugin author.  the current behavior of overriding initialize is kinda ugly.


provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
```diff
@@ -21,14 +21,18 @@
...
-    super(task)
+  def initialize(env, task)
+    super(env, task)
+    work_dir = @env[:work_dir]
+    tenant = @env[:tenant]
     # local and remote top-level script directory name
     @scripts_parent_dir = "scripts"
     # local scripts dir
-    @scripts_dir = "#{File.expand_path(File.dirname(__FILE__))}/#{@scripts_parent_dir}"
+    #@scripts_dir = "#{File.expand_path(File.dirname(__FILE__))}/#{@scripts_parent_dir}"
```
albertshau: remove comment

provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
```diff
@@ -21,14 +21,18 @@
...
+    @scripts_dir = %W[ #{work_dir} #{tenant} automatortypes shell #{@scripts_parent_dir} ].join('/')
     # name of tarball to generate
-    @scripts_tar = "#{File.expand_path(File.dirname(__FILE__))}/scripts.tar.gz"
+    #@scripts_tar = "#{File.expand_path(File.dirname(__FILE__))}/scripts.tar.gz"
```
albertshau: comment

albertshau:
lgtm

PR #369, opened 2014-08-13
Feature/standalone data directories

dereklwood:
additional standalone config to keep all data in its local `./data` dir instead of /var/loom...
- [x] added `server.plugin.store.localfilestore.data.dir` to server config template
- [x] applied the startup sed to the `provisioner-site.xml` as well, and added data/work dir settings to it


albertshau:
lgtm

PR #373, opened 2014-08-13
Feature/ui add sshuser

gaarf:
- adds `sshuser` field to imagetype provider form ui.
- refactored corresponding JS to be more readable.

Fixes #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Continuuity Loom CHANGELOG
 
 Unreleased
 ----------
+- UI: Add 'sshuser' field to imagetypes ( Issues: #366 )
 - AWS provider support ( Issues: #22 #322 )
 - MongoDB support for clusters ( Issues: #43 #130 )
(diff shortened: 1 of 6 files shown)


wolf31o2:
Tests need to be updated. Also, you should update CHANGELOG.md with this and #366 and push it to this branch, too.

gaarf:
@wolf31o2 it appears that the failure happens if UI submits a `providermap` with a `sshuser` key set to an empty string. Should the UI enforce presence of a sshuser string, or is there a default value?

wolf31o2:
The UI should not enforce presence. We default them on the provisioner side. We _could_ default them on the UI to "root" but we've kept that sort of logic out of the UI, previously.

gaarf:
@wolf31o2 actually the test failure happens no matter what the value of `sshuser`. If the key exists, the API responds with error. Are you sure backend is ready or am I missing something?

wolf31o2:
I am sure that the backend is ready, as I am using this functionality (storing data to server via REST and provisioning a cluster).

wolf31o2:
:+1: LGTM...

PR #374, opened 2014-08-13
Import hadoop_wrapper version 0.1.6

wolf31o2:
This fixes an issue with using Hive on singlenode clustertemplates and allows setting our YARN memory allocation via a percentage (instead of absolute)...


albertshau:
lgtm

PR #375, opened 2014-08-14
add rest_client requirement

dereklwood:
provisioner requires rest_client gem, which used to be pulled in by one of the other deps


gaarf:
:thumbsup:

PR #379, opened 2014-08-15
code cleanup. Removing unused stepSize and ratio from service

albertshau:
constraints and removing strings that get turned into json that
get turned into objects in favor of just the objects.


poornachandra:
LGTM :+1:

PR #380, opened 2014-08-15
Feature/update kerberos

wolf31o2:
Update the `krb5` and `krb5_utils` cookbooks to the latest versions.


wolf31o2:
OTS via @dereklwood

PR #382, opened 2014-08-15
Feature/data uploader utility

dereklwood:
- [x] data-uploader utility to upload a single resource to loom server.
- [x] does a lot of pre-validation with discovery of expected format, etc
- [ ] the rest-client library is supposed to upload files automatically as multipart, but doesn't look like its working, so i may need to replace with something else.  but it should work for all our stuff.
- [x] it does the gzipping in pure ruby but also in-memory for now.  If needed it can upload a pre-existing tgz.
- [x] updated standalone.xml so data-uploader.rb is packaged into `${standalone-root}/bin`

```
$ ./data-uploader.rb -h
Usage: ./data-uploader.rb [options] <action> <local-path> <remote-target>
    -u, --uri URI                    Server URI, defaults to ENV['LOOM_SERVER_URI'] else "http://localhost:55054"
    -t, --tenant TENANT              Tenant, defaults to ENV['LOOM_TENANT'] else "superadmin"
    -U, --user USER                  User, defaults to ENV['LOOM_API_USER'] else "admin"

Required Arguments:
         <action>: one of upload, stage, or sync (sync acts on all staged resources, so can be dangerous)
     <local-path>: path to the local copy of the resource to upload
  <remote-target>: api path defining the resource

Example:
  ./data-uploader.rb -u http://localhost:55054 -t superadmin -U admin sync ./my/local/cookbooks/hadoop automatortypes/chef-solo/cookbooks/hadoop

```

```
./data-uploader.rb -u http://localhost:55054 -t superadmin -U admin sync /Users/derek/git/hadoop_cookbook automatortypes/chef-solo/cookbooks/hadoop
upload successful for http://localhost:55054/v1/loom/automatortypes/chef-solo/cookbooks/hadoop, version: 15
stage successful for http://localhost:55054/v1/loom/automatortypes/chef-solo/cookbooks/hadoop/versions/15/stage
sync successful
```


dereklwood:
created #384 to fix memory usage

bin/data-uploader.rb
```diff
@@ -0,0 +1,303 @@
...
+  opts.on('-U', '--user USER', 'User, defaults to ENV[\'LOOM_API_USER\'] else "admin"') do |u|
+    options[:user] = u
+  end
+  opts.on('-q', '--quiet', 'Suppress all non-error output') do
+    options[:quiet] = true
+  end
+
+  opts.separator ''
+  opts.separator 'Required Arguments:'
+  opts.separator '         <action>: one of upload, stage, or sync (sync acts on all staged resources, so can be dangerous)'
```
albertshau: should make it clear that stage will upload and stage at the same time, and that sync will upload, stage, and sync all at once.

bin/data-uploader.rb
```diff
@@ -0,0 +1,303 @@
...
+options = {}
+OptionParser.new do |opts|
+  opts.banner = "Usage: #{$PROGRAM_NAME} [options] <action> <local-path> <remote-target>"
+  opts.on('-u', '--uri URI', 'Server URI, defaults to ENV[\'LOOM_SERVER_URI\'] else "http://localhost:55054"') do |u|
+    options[:uri] = u
+  end
+  opts.on('-t', '--tenant TENANT', 'Tenant, defaults to ENV[\'LOOM_TENANT\'] else "superadmin"') do |t|
+    options[:tenant] = t
+  end
+  opts.on('-U', '--user USER', 'User, defaults to ENV[\'LOOM_API_USER\'] else "admin"') do |u|
```
albertshau, line 40: probably not needed, only admins are allowed to use these apis.

albertshau:
minor comments, otherwise lgtm

bin/data-uploader.rb
dereklwood, line 40: after discussion, going to leave it in.  leaving it out will be a bit awkward, and it may still be used when we do the security stuff

PR #386, opened 2014-08-15
Feature/standalone sync bundled data

dereklwood:
depends on #382 
- [x] loom.sh standalone startup script will now source any `load-bundled-data.sh` included at plugin top-level.  these scripts can use the common data-uploader.rb to do the actual work.  
- [x] load-bundled-data.sh included here for chef-solo, shell coming next
- [x] loom.sh waits and confirms plugin registration, stages all data, then does a sync

```
$ time ./loom.sh start
Starting Loom Server ...
Starting Loom UI ...
Waiting for server to start before loading default configuration...
Loading default configuration...
Registering provisioner plugins with configured server
Waiting for plugins to be registered...
Loading initial data...
Syncing initial data...
Waiting for server to start before running provisioner...
Starting Loom Provisioner ...
Requesting 5 workers for default tenant...

Go to http://localhost:8100. Have fun creating clusters!

real    0m32.520s
user    0m17.531s
sys 0m2.270s
```

### DIFF
--- a/provisioner/worker/plugins/automators/chef_solo_automator/load-bundled-data.sh
+++ b/provisioner/worker/plugins/automators/chef_solo_automator/load-bundled-data.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+TIMEOUT=3
+LOOM_HOME=${LOOM_HOME:-/opt/loom}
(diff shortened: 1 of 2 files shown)


albertshau:
lgtm

PR #389, opened 2014-08-18
Feature/data tar fix

dereklwood:
fix issues related to tarring and untarring.
- [x] the tar creation in data-uploader was incorrectly setting the parent dir on entries, and was also not creating an explicit entry for just this parent dir (ie, top-level 'hadoop')
- [x] in the provisioner, attempt to workaround this problem if it ever happens (shouldn't)


albertshau:
lgtm

PR #393, opened 2014-08-19
Feature/build php modules

wolf31o2:
Fixes #355

### DIFF
--- a/config/defaults/clustertemplates/lamp.json
+++ b/config/defaults/clustertemplates/lamp.json
@@ -49,8 +49,7 @@
           "mysql.default_host": "%host.service.mysql-server%"
         },
         "modules": [
-          "memcache",
-          "mysql"
+          "memcache"
         ]
       }
     }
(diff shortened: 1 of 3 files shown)


config/defaults/clustertemplates/lamp.json
```diff
@@ -49,8 +49,7 @@
           "mysql.default_host": "%host.service.mysql-server%"
         },
         "modules": [
-          "memcache",
-          "mysql"
```
albertshau, line 53: how come the mysql module is getting removed?

albertshau:
one question, otherwise lgtm

config/defaults/clustertemplates/lamp.json
wolf31o2, line 53: Second commit message: "Remove mysql, as it appears that it has been merged into the main PHP package"

There is no longer a mysql module for PHP. The older mysql and mysqli modules have been removed from PEAR/PECL due to the inclusion in mainline.

PR #397, opened 2014-08-20
changing test to post one node at a time to be in line with the

albertshau:
api change.


seawatts:
Sorry about that. I had fixed it locally but forgot to push back up

seawatts:
LGTM

PR #400, opened 2014-08-21
Feature/shell provisioner data

dereklwood:
this PR updates the shell provisioner in the current multitenant world, feature parity with what we had before.
- [x] provides resource types scripts (file) and archives (archive).  scripts is essentially what we had before, uploadable one file at a time, synced to scripts/[filename].  archives is a way to upload directories at a time... they are accessible namespaced under archives/[dir-name]/...
- [x] minor updates to the json parsing stuff to continue working.  `.lib` is now just a top-level `lib`
- [x] minor refactoring now that shell provisioner processes multiple tar files (files, archives, lib)
- [ ] identified a major gap with file permissions which we will need to address separately


provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
```diff
@@ -25,28 +25,46 @@ def initialize(env, task)
...
-    @remote_cache_dir = "/var/cache/loom"
+    @remote_cache_dir = "/var/cache/loom/shell_automator"
     # remote script location to be exported in $PATH
-    @remote_scripts_dir = "#{@remote_cache_dir}/#{@scripts_parent_dir}"
+    @remote_scripts_dir = "#{@remote_cache_dir}/scripts"
+    # remote lib location
+    @remote_lib_dir = "#{@remote_cache_dir}/#{@lib_parent_dir}"
     # loom wrapper for common functions
-    @wrapper_script = "#{@remote_scripts_dir}/.lib/loom_wrapper.sh"
+    @wrapper_script = "#{@remote_lib_dir}/loom_wrapper.sh"
```
albertshau, line 46: since we're renaming anyway, can we remove 'loom' from the name of the script

albertshau, line 46: eh that might affect too many services, lets do it in a different PR

provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
```diff
@@ -25,28 +25,46 @@ def initialize(env, task)
...
+  end
+
+  def generate_resource_tar(resource)
+    resource_path = %W[ #{@resources_path} #{resource} ].join('/')
+    resource_tar  = %W[ #{@resources_path} #{resource}.tar.gz ].join('/')
+    if !File.exist?(resource_tar) or ((Time.now - File.stat(resource_tar).mtime).to_i > 600)
```
albertshau: this is very similar to generate_lib_tar. We can create a single generate_tar function that both of them call.

provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
```diff
@@ -25,28 +25,46 @@ def initialize(env, task)
...
-  def generate_scripts_tar
-    if !File.exist?(@scripts_tar) or ((Time.now - File.stat(@scripts_tar).mtime).to_i > 600)
-      log.debug "Generating #{@scripts_tar} from #{@scripts_dir}"
-      scripts_tar_path = File.dirname(@scripts_dir)
-      scripts_parent_dir = File.basename(@scripts_dir)
-      `tar -cLzf "#{@scripts_tar}.new" -C "#{scripts_tar_path}" #{scripts_parent_dir}`
-      `mv "#{@scripts_tar}.new" "#{@scripts_tar}"`
-      log.debug "Generation complete: #{@scripts_tar}"
+  def generate_lib_tar
+    if !File.exist?(@lib_tar) or ((Time.now - File.stat(@lib_tar).mtime).to_i > 600)
```
albertshau: would be good to have the 10 minutes be a config setting, though dont have to do it now.

albertshau:
small comment, otherwise lgtm

provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
dereklwood: generate_resource_tar is the generic one to handle any defined resource, while generate_lib_tar is a specific need of the plugin itself... so they have different inputs.  I considered combining them but put it off.  i'll take another look if you feel strongly about it.

provisioner/worker/plugins/automators/shell_automator/shell_automator.rb
dereklwood: agreed, though right now i'd have to pass it as a cmdline argument from the master to the worker, which is a bit ugly.  I am planning to more tightly integrate the master and worker into the same project/gem so they can share the same config file and utilities.

PR #402, opened 2014-08-21
dnssuffix field in createcluster advanced section 

gaarf:
fixes #52

the field will be hidden in the cluster-reconfigure view.


albertshau:
:+1:

PR #413, opened 2014-08-22
Feature/aws disks raid

wolf31o2:
- [x] Detect instance-store ephemeral disks
- [x] Install mdadm (requires a configured mailer on Ubuntu, so use nullmailer)
- [x] Create RAID, if enough disks
- [x] Mount /data disk

This is the instance-store part of issue #405


dereklwood:
looks like a good reasonable default for now.  :+1:

PR #415, opened 2014-08-22
Feature/organize specifications

albertshau:


server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
```diff
@@ -24,11 +24,8 @@
 import com.continuuity.loom.store.tenant.TenantStore;
 import com.google.common.base.Charsets;
 import com.google.gson.Gson;
-import com.google.gson.JsonArray;
 import com.google.gson.JsonIOException;
-import com.google.gson.JsonObject;
 import com.google.gson.JsonSyntaxException;
-import com.google.gson.reflect.TypeToken;
```
wolf31o2, line 31: Assuming IDE tells you to remove things like this when they're unused?

wolf31o2:
:+1: Other than a question about changes that aren't s/admin/spec/ it all LGTM...

server/src/main/java/com/continuuity/loom/http/handler/LoomNodeHandler.java
albertshau, line 31: yep

PR #416, opened 2014-08-22
Feature/joyent certified ubuntu

wolf31o2:
Switch joyent to using the certified Ubuntu images. This required cleaning up privileged calls for non-root users and also standardizing data disk mounting to `/data` to match Joyent's other images.

Fixes #315


albertshau:
lgtm assuming its been tested

PR #417, opened 2014-08-22
Feature/use data dir

wolf31o2:
Most providers give us secondary data disks, which we are mounting to `/data`. This tells our templates to use this location for data.


albertshau:
:+1:

PR #420, opened 2014-08-25
ui cluster reconfig / provider fixes

gaarf:
Fixes for issues #385, #321 and #210

### DIFF
--- a/ui/client/js/user/createcluster.js
+++ b/ui/client/js/user/createcluster.js
@@ -234,7 +234,9 @@ CreateCluster.app.controller('CreateClusterCtrl', ['$scope', '$interval', 'dataF
 CreateCluster.populateDefaults = function (providerInfo, data) {
   angular.forEach(data.parameters, function(o) {
     angular.forEach(o.fields, function(v, name) {
-      providerInfo.provisioner[name] = v.default || ''; 
+      if(v.default) {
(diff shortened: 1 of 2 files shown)


ui/client/js/user/createcluster.js
```diff
@@ -114,8 +114,12 @@ CreateCluster.app.controller('CreateClusterCtrl', ['$scope', '$interval', 'dataF
   $scope.$watch('defaultProvider', function () {
     if ($scope.defaultProvider) {
       dataFactory.getProviderFields($scope.defaultProvider, function (providerInfo) {
-        $scope.providerFields = $scope.providerData[providerInfo.providertype];
-        $scope.defaultProviderInfo = providerInfo;
+        var data = $scope.providerData[providerInfo.providertype];
+        $scope.providerFields = data;
+
+        if(!$scope.clusterId) {
+          $scope.defaultProviderInfo = CreateCluster.populateDefaults(providerInfo, data);
```
albertshau, line 121: It seems like this overwrites fields that already have values.  For example, if I create a cluster with the rackspace provider, the value for 'region' should be the one the admin has set, which may differ from the default.

albertshau:
one comment, otherwise lgtm

albertshau:
:+1:

PR #423, opened 2014-08-25
protect against nil permissions

dereklwood:
fixes #422 
- [x] stop introducing nil permissions in resourcespec
- [x] resourcemanager check again for nil before using


albertshau:
lgtm

wolf31o2:
:+1: LGTM

PR #430, opened 2014-08-27
Adding cluster progress to the clusters get call, as well as

albertshau:
the cluster provider and template as objects instead of strings.
This is to preserve the structure of these objects when a get
call is made for a specific cluster to keep the apis restful.
Also using java objects instead of manipulating json objects
for those.

Including progress requires adding another column to the
clusters table in order to allow joining on the job number. Since
this makes the getClusterStatuses rpc call useless, removed it.


albertshau:
fixes issue #401

docs/docs/source/rest/clusters.rst
```diff
@@ -250,17 +318,52 @@ To get the status of a cluster, make a GET HTTP request to URI:
 Status of a cluster is a JSON Object with a clusterid, stepstotal, stepscompleted, 
 status, actionstatus, and action.  
 
-The status can be one of PENDING, ACTIVE, INCOMPLETE,
-and TERMINATED.  PENDING means there is some actions pending, ACTIVE means the cluster 
-is active and can be used, INCOMPLETE means there was some previous action failure so 
-the cluster may not be usable, but is deletable, and TERMINATED means the cluster is 
+The status can be one of pending, active, incomplete, and terminated.
+pending means there is some actions pending, active means the cluster
```
wolf31o2: Sentence starts lowercase...

server/src/main/java/com/continuuity/loom/store/cluster/SQLSystemClusterStoreView.java
```diff
@@ -87,6 +89,14 @@ protected PreparedStatement getSelectClusterJobsStatement(Connection conn, long
   }
 
   @Override
+  PreparedStatement getSelectAllClusterJobsStatement(Connection conn) throws SQLException {
+    PreparedStatement statement = conn.prepareStatement(
+      "SELECT C.cluster, J.job FROM clusters C, jobs J " +
+        "WHERE C.latest_job_num=J.job_num AND C.id=J.cluster_id ORDER BY C.create_time DESC");
```
wolf31o2, line 95: Is it necessary to ORDER BY for this?

server/src/main/java/com/continuuity/loom/store/cluster/SQLUserClusterStoreView.java
```diff
@@ -103,8 +107,18 @@ protected PreparedStatement getSelectClusterJobsStatement(Connection conn, long
...
+    statement.setString(2, tenantId);
+    statement.setString(3, userId);
+    return statement;
+  }
+
+  @Override
+  PreparedStatement getSelectAllClusterJobsStatement(Connection conn) throws SQLException {
+    PreparedStatement statement = conn.prepareStatement(
+      "SELECT C.cluster, J.job FROM clusters C, jobs J WHERE C.latest_job_num=J.job_num AND C.id=J.cluster_id " +
+        "AND C.tenant_id=? AND C.owner_id=? ORDER BY C.create_time DESC");
```
wolf31o2, line 119: Again with the ORDER BY...

wolf31o2:
Minor SQL comments, otherwise, LGTM...

server/src/main/java/com/continuuity/loom/store/cluster/SQLSystemClusterStoreView.java
albertshau, line 95: technically no, but the previous behavior ordered so wanted to be consistent.

wolf31o2, line 95: I guess it's fine. I prefer not ordering simply because it incurs a performance overhead and isn't necessary. That being said, it shouldn't prevent this from being merged.

PR #433, opened 2014-08-27
Feature/pngcrush

wolf31o2:
This changes over 3000 files, but there should be no content change.

Ran the following:

``` bash
find . -name '*.png' -exec pngcrush -reduce -fix -new -ow {} \;
```

I did _not_ pngcrush any files shipped from third parties, so nothing under `ui/node_modules` or `ui/client/third_party` has been crushed.

Results:

```
$ git checkout develop && du -sch . | grep total
679M    total

$ git checkout feature/pngcrush && du -sch . | grep total
652M    total
```


albertshau:
is this necessary? almost certainly doesn't break anything but no way to really know for sure besides looking at everything.

wolf31o2:
This is purely a bandwidth-saver... it isn't necessary. As for checking the images, we could do that, but I have a good bit of experience with pngcrush. It only updates images when it's able to do a conversion, like down-sampling the bit depth when you're only using "basic" colors.

wolf31o2:
I am going to close this. We can, instead, crush the new screenshots for 0.9.8

PR #434, opened 2014-08-27
Feature/update templates

wolf31o2:
Minor cleanups to `hadoop-distributed` and lots of memory-saving options on `hadoop-singlenode` template.


config/defaults/clustertemplates/hadoop-distributed.json
```diff
@@ -96,7 +97,7 @@
   },
   "constraints": {
     "size": {
-      "min": 3
+      "min": "3"
```
albertshau, line 100: this doesn't have to be a string. doesn't matter either way, but its ok to be a number.

albertshau:
lgtm

config/defaults/clustertemplates/hadoop-distributed.json
wolf31o2, line 100: I know. I tend to make them strings because Ruby autotypes stuff and a lot of code assumes a string and not an integer. We've managed to check and convert types when necessary, but I tend to be cautious.

PR #437, opened 2014-08-28
Server expects the provisioner fields in a hardware type to be

albertshau:
a flat map of string to string, so flattening the hardware type
informational fields.


wolf31o2:
The disk section is a bit hard to use. Previously, it was easy to use `disks.each do |n, v|` to get the optional disks with n = name and v = size. This requires more complex parsing. I would prefer something more like:

``` json
{
  "joyent": {
    "disks": "/dev/vdb=420,/dev/vdc=420"
  }
}
```

wolf31o2:
:+1: LGTM

PR #438, opened 2014-08-28
Allow provisioners to overwrite the server given hostname for a

albertshau:
node, just like how they can set the ip addresses for a node.


wolf31o2:
:+1: LGTM, but not mergable

PR #442, opened 2014-08-28
Feature/add credential store

albertshau:


wolf31o2:
Build failure. Also, does this require an external memcached, and do we have some in-memory version when external memcached is not available?

albertshau:
does not require an external memcached, uses server memory by default.

wolf31o2:
Ahh, right. You told me that. Well, it LGTM... assuming build completes

server/src/main/java/com/continuuity/loom/scheduler/task/ClusterService.java
```diff
@@ -574,6 +599,19 @@ private void validateAndAddFieldsToProvider(Provider provider, ClusterCreateRequ
     }
 
     // check all required user fields are present
+    // if there are no fields in the request, they may be in the credential store
+    if (providerFields.isEmpty()) {
+      try {
+        Map<String, String> existingSensitiveFields = credentialStore.get(account.getTenantId(), clusterId);
```
wolf31o2, line 605: What if the field isn't a string, such as when it's a boolean? Do we convert everything to string?

albertshau, line 605: Yep.  You bring up a good point though.  The original spec for providers have all their fields as strings, which is inconsistent with the types we allow plugin authors to set today. Created Issue #444

PR #443, opened 2014-08-28
fix typos in javadocs.

albertshau:


server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -164,7 +164,7 @@ public void writeTenantSpecification(TenantSpecification tenantSpecification)
    * @param tenantId Id of the tenant to verify quotas for
    * @param additionalClusters Number of clusters that would be added
    * @param additionalNodes Number of nodes that would be added
-   * @returns True if the tenantQuotas would be satisfied, false if they would be exceeded.
+   * @return True if the tenant quotas would be satisfied, false if they would be exceeded.
```
awholegunch: I'm thinking "true" should be lowercase, as in
- @return true if the tenant quotas would be satisfied, false if they would be exceeded.

server/src/main/java/com/continuuity/loom/provisioner/TenantProvisionerService.java
```diff
@@ -183,7 +183,7 @@ public boolean satisfiesTenantQuotas(String tenantId, int additionalClusters,
    * @param tenant Tenant to verify quotas for
    * @param additionalClusters Number of clusters that would be added
    * @param additionalNodes Number of nodes that would be added
-   * @returns True if the tenantQuotas would be satisfied, false if they would be exceeded.
+   * @return True if the tenant quotas would be satisfied, false if they would be exceeded.
```
awholegunch: see above.

PR #448, opened 2014-08-29
Feature/rubocop todo

wolf31o2:
This is nothing but rubocop updates... Rakefile now outputs the culprit cops when one fails, added .rubocop.yml to plugins, and updated .rubocop_todo.yml files.


dereklwood:
:+1:

PR #456, opened 2014-08-30
Feature/hostsfile access v4

wolf31o2:
Add hosts entries for access_v4 and bind_v4 ipaddresses and don't enforce uniqueness on bind_v4


dereklwood:
:+1:

PR #457, opened 2014-08-30
Add .gitignore for scripts/validation.pem

wolf31o2:
This allows us to store our validation.pem in the scripts directory for chef_client_bootstrap.sh without committing it to the repository.


dereklwood:
as discussed, the 755 perms should be fine in this case. :+1:

PR #460, opened 2014-09-02
adding encryption capabilities for encryping data sent to the

albertshau:
credential store.


coopr-server/src/main/java/co/cask/coopr/store/credential/MemcachedCredentialStore.java
```diff
@@ -58,12 +66,24 @@ private MemcachedCredentialStore(Configuration conf) throws IOException {
     this.ttlSeconds = ttlSeconds;
     this.timeoutSeconds = conf.getInt(Constants.MemcachedCredentialStore.TIMEOUT,
                                       Constants.MemcachedCredentialStore.DEFAULT_TIMEOUT);
+
+    CipherProvider cipherProvider = CipherProvider.builder()
+      .setTransformation(conf.get(Constants.MemcachedCredentialStore.ENCRYPT_TRANSFORMATION))
```
poornachandra, line 71: If the constants are not dependent on memcached then just call them Constants.CredentialStore

poornachandra:
As we discussed, it would be a good idea to store sensitive configuration in a different configuration file so that you can have more restrictive permissions on it. That can be done in a separate PR. 
Otherwise LGTM :+1:

coopr-server/src/main/java/co/cask/coopr/store/credential/MemcachedCredentialStore.java
albertshau, line 71: right now it is inside this class, but it probably makes sense to invert the relationship. Can do that separately.

PR #463, opened 2014-09-02
Feature/bind to self

wolf31o2:
Force NameNode to use it's bind_v4 address, by explicit IP, rather than resolving for it


albertshau:
lgtm

PR #464, opened 2014-09-02
Import hadoop version 1.3.2

wolf31o2:
This fixes a bug in the `zookeeper` recipe which affects certain Hadoop distributions.


albertshau:
:+1:

PR #470, opened 2014-09-03
default the user if the field is empty as well.

albertshau:


dereklwood:
:+1: lgtm

server/src/main/java/com/continuuity/loom/cluster/NodeProperties.java
```diff
@@ -56,7 +56,7 @@ private NodeProperties(String hostname, Map<String, String> ipaddresses, int nod
     this.flavor = flavor;
     this.image = image;
     this.nodenum = nodenum;
-    this.sshuser = sshUser == null ? "root" : sshUser;
+    this.sshuser = sshUser == null || sshUser.isEmpty() ? "root" : sshUser;
```
gaarf: it would be nice to use some parens here, to make it obvious wether the OR has precedence over the ternary operator.

`this.sshuser = (sshUser == null || sshUser.isEmpty()) ? "root" : sshUser;`

PR #473, opened 2014-09-04
Feature/firewall updates

wolf31o2:
This adds firewall entries for both `access_v4` and `bind_v4` addresses, if they're unique. Otherwise, it adds an entry for `access_v4`...

Tested on AWS (unique addresses) and Rackspace (identical addresses)...


dereklwood:
:+1: lgtm

PR #474, opened 2014-09-04
Feature/google support

dereklwood:
GCE support.  some notes:
- disk creation/confirmation happens first in the create task
- a data disk of configurable size is mounted to /data if given
- requiretty sudoers config is disabled if present
- selinux is set to permissive if present
- provisioner requires an ssh user/key, which must be configured in the GCE project metadata.  the user can be defined in the plugin ssh_username field (optional, takes precedence), or else defaults to the image/server value
- GCE default network is firewalled externally on all ports except 22 by default.  some initial firewall configuration or custom network definition outside of loom will likely be needed.
- the ability to disable ephemeral IP assignment will require fog upgrade to 1.23.  will do in a future PR.


provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -121,6 +121,85 @@
...
+              "us-central1-a",
+              "us-central1-b",
+              "us-central1-f"
+            ],
+            "default": "us-central1-a",
+            "override": true,
+            "tip": "Zone"
+          }
+        },
+        "required": [
```
wolf31o2, line 180: ssh_username is not required?

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -121,6 +121,85 @@
...
+            "key_location",
+            "zone_name",
+            "ssh_keyfile"
+          ]
+        ]
+      },
+      "user": {
+        "fields": {
+          "data_disk_size_gb": {
+            "label": "Data disk size in Gb",
```
wolf31o2: GB

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+  include FogProvider
+
+  def create(inputmap)
+    @flavor = inputmap['flavor']
+    @image = inputmap['image']
+    @hostname = inputmap['hostname']
+    # GCE does not allow dots, including the loom-server-appended .local
+    # we generate the unique providerid from the loom-hostname
+    # and leave the loom hostname for use in /etc/hosts
+    @providerid = @hostname[/[a-zA-Z0-9\-]*/]
```
wolf31o2: This is a little confusing, IMO. I would have used:

``` ruby
@providerid = @hostname.split('.').first
```

Are there restrictions on the providerid that are different than that of a hostname?

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+    fields = inputmap['fields']
+    begin
+      # set instance variables from our fields
+      fields.each do |k, v|
+        instance_variable_set('@' + k, v)
+      end
+      # validate credentials
+      validate!
+      # Create the server
+      log.debug "Creating #{@providerid} on GCE using flavor: #{flavor}, image: #{image}"
```
wolf31o2: Do we call it "google"/"Google" or "google"/"Google Compute" or "gce"/"GCE"? Let's pick something and be consistent, at least in anything outward facing.

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
dereklwood, line 180: to support the use case of defining username in the image like the other providers

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      # create the VM
+      connection.servers.create(create_server_def)
+
+      # Process results
+      # return the unique providerid we used
+      @result['result']['providerid'] = @providerid
+      # set ssh user
+      ssh_user =
+        if @ssh_username.to_s != ''
```
wolf31o2: These should already be strings, when received from the server.

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: it's equivalent... google's client returns that regex in the error message, so thought it best to use it.  but i guess its the same as the hostname scrubbing the server should be doing... so either way

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+      fields.each do |k, v|
+        instance_variable_set('@' + k, v)
+      end
+      # validate credentials
+      validate!
+      # Confirm server
+      log.debug "Invoking server confirm for id: #{providerid}"
+      server = connection.servers.get(providerid)
+      # Wait until the server is ready
+      fail 'Server #{server.name} is in ERROR state' if server.state == 'ERROR'
```
wolf31o2: Double quotes, or it'll display `#{server.name}` instead of the server name. Also, does GCE have a server.name?

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      # do we need sudo bash?
+      sudo = 'sudo' unless @task['config']['ssh-auth']['user'] == 'root'
+
+      set_credentials(@task['config']['ssh-auth'])
+
+      # login with pseudotty and turn off sudo requiretty option
+      log.debug "attempting to ssh to #{bootstrap_ip} as #{@task['config']['ssh-auth']['user']} with credentials: #{@credentials} and pseudotty"
+      Net::SSH.start(bootstrap_ip, @task['config']['ssh-auth']['user'], @credentials) do |ssh|
+        cmd = %Q[#{sudo} cat /etc/sudoers | sed 's/^\\(Defaults\\s\\+requiretty.*\\)$/#\\1/i' > /tmp/sudoers.new && #{sudo} visudo -c -f /tmp/sudoers.new && #{sudo} EDITOR="cp /tmp/sudoers.new" visudo && rm -f /tmp/sudoers.new]
```
wolf31o2: Why not `#{sudo} sed -i` ?

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+        'bind_v4' => bind_ip
+      }
+
+      # do we need sudo bash?
+      sudo = 'sudo' unless @task['config']['ssh-auth']['user'] == 'root'
+
+      set_credentials(@task['config']['ssh-auth'])
+
+      # login with pseudotty and turn off sudo requiretty option
+      log.debug "attempting to ssh to #{bootstrap_ip} as #{@task['config']['ssh-auth']['user']} with credentials: #{@credentials} and pseudotty"
```
wolf31o2: Attempting

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+      Net::SSH.start(bootstrap_ip, @task['config']['ssh-auth']['user'], @credentials) do |ssh|
+        cmd = %Q[#{sudo} cat /etc/sudoers | sed 's/^\\(Defaults\\s\\+requiretty.*\\)$/#\\1/i' > /tmp/sudoers.new && #{sudo} visudo -c -f /tmp/sudoers.new && #{sudo} EDITOR="cp /tmp/sudoers.new" visudo && rm -f /tmp/sudoers.new]
+        ssh_exec!(ssh, cmd, 'disabling requiretty via pseudotty session', true)
```
wolf31o2: Disabling

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: note, this is also broken in aws where i copied it from.  and yes, GCE server name is the short-hostname

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: this is to check for nil or empty string (nil.to_s = '')

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+        next if disk.key?('boot') && disk['boot'] == true
+        # fog attaches additional disks as 'persistent-disk-[index]', google prepends 'google-'
+        if disk.key?('deviceName') && disk['deviceName'] =~ /^persistent-disk-(\d+)/
+          mount_point = "/data#{Regexp.last_match[1]}"
+          mount_point = '/data' if mount_point == '/data1'
+          google_disk_id = "google-#{disk['deviceName']}"
+
+          # Mount the data disk
+          Net::SSH.start(bootstrap_ip, @task['config']['ssh-auth']['user'], @credentials) do |ssh|
+            cmd = %Q[#{sudo} mkdir #{mount_point} && #{sudo} /usr/share/google/safe_format_and_mount -m 'mkfs.ext4 -F' /dev/$(basename $(readlink /dev/disk/by-id/#{google_disk_id})) #{mount_point} && #{sudo} chmod a+w #{mount_point}]
```
wolf31o2: Fetch your device name to a variable prior to running this SSH command, so you don't have to sub-shell to readlink multiple times.

Also, why the chmod a+w to make it world-writable?

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+            ssh_exec!(ssh, cmd, "mounting device #{google_disk_id} on #{mount_point}")
```
wolf31o2: Mounting

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+            # update /etc/fstab
+            cmd = %Q[echo "/dev/$(basename $(readlink /dev/disk/by-id/#{google_disk_id})) #{mount_point} ext4 defaults,auto,noatime 0 2" | #{sudo} tee -a /etc/fstab]
+            ssh_exec!(ssh, cmd, "updating fstab for device #{google_disk_id} on #{mount_point}")
```
wolf31o2: Updating

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+          end
+        else
+          log.warn "unexpected disk device found, ignoring: #{disk}"
```
wolf31o2, line 172: Unexpected

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+        end
+      end
+
+      # disable SELinux
+      Net::SSH.start(bootstrap_ip, @task['config']['ssh-auth']['user'], @credentials) do |ssh|
+        cmd = "if [ -x /usr/sbin/sestatus ] ; then #{sudo} /usr/sbin/sestatus | grep disabled || ( test -x /usr/sbin/setenforce && #{sudo} /usr/sbin/setenforce Permissive ) ; fi"
```
wolf31o2: `s/[/test/` since `[` is a shell built-in (some distributions, like Gentoo symlink `/bin/[` to `/bin/test` for this reason)

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+      # delete server
+      log.debug 'Invoking server delete'
+      server = connection.servers.get(providerid)
+      disks = server.disks
+      begin
+        server.destroy
+        server.wait_for(120) { !ready? }
+      rescue Fog::Errors::NotFound
+        # ok, can be thrown by wait_for
+        log.debug 'disk no longer found'
```
wolf31o2: This doesn't make sense. Is this supposed to be server?

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+      end
+      # delete all attached disks
+      disks.each do |d|
+        name = d['source'].split('/')[-1]
```
wolf31o2: Use .last instead of array position as a number.

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+        disk = connection.disks.get(name)
+        log.debug "deleting disk #{name}"
+        begin
+          disk.destroy
+          disk.wait_for(120) { !ready? }
```
wolf31o2, line 244: Is it necessary to wait_for or can we fire off the delete request and continue to the next disk?

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -121,6 +121,85 @@
       }
     }
   },
+  "google": {
+    "name": "google",
+    "classname": "FogProviderGoogle",
+    "parameters": {
+      "admin": {
+        "fields": {
+          "project_id": {
```
wolf31o2: All other providers use a provider-specific prefix... google_project_id

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,297 @@
...
+
+  def connection
+    # Create connection
+    # rubocop:disable UselessAssignment
+    @connection ||= begin
+      connection = Fog::Compute.new(
+        provider: 'google',
+        google_project: @project_id,
+        google_client_email: @client_email,
+        google_key_location: @key_location
```
wolf31o2: In most providers, where possible, I have tried to keep the JSON variable the same as the fog one. Should we be using `google_project`, `google_client_email`, and `google_key_location` instead? I think so.

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
wolf31o2, line 180: Got it. Since using a key from metadata will add the user to the machine, even if it's not defined in the image, correct?

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -121,6 +121,85 @@
...
+            "client_email",
+            "key_location",
+            "zone_name",
+            "ssh_keyfile"
+          ]
+        ]
+      },
+      "user": {
+        "fields": {
+          "data_disk_size_gb": {
```
wolf31o2: Move to admin, per our discussion.

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,296 @@
...
+      disks = server.disks
+      begin
+        server.destroy
+        server.wait_for(120) { !ready? }
+      rescue Fog::Errors::NotFound
+        # ok, can be thrown by wait_for
+        log.debug 'disk no longer found'
+      end
+      # delete all attached disks
+      disks.each do |d|
```
albertshau: What happens if the server destroy succeeds but the disk destroys fail? When Loom retries the delete, and   connection.servers.get(providerid) gets run again, will it return the deleted server with the disks?  If not we should delete the disks before deleting the server.

albertshau:
1 functionality comment, will leave ruby to chris

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: this is the disk leak issue i brought up and we decided against it for the reason that someone can detach disks outside of loom management.  disks can not be deleted before the server.  one possible solution is to attempt to detach first but i still need to store all this disk state info in the payload, which is ugly (we need a higher level management layer).

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: standardized on Google

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: safety, originally, but i agree simplicity is better in this case.  fixed

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: fixed, and removed chmod... letting automators deal with it

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: fixed

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: typo, fixed

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: nothing wrong with -1 but changed it anyway

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood, line 244: reworked this to do it in parallel.  it only gets us minimal gain for now... maybe if we allow more disks.  after revisiting this, i think it is worth storing the disk metadata in the payload so that delete can be retried effectively for server and/or disk.  this probably warrants its own PR though

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
dereklwood: aligned fields with fog vars, and prepended google_ to non-fog vars

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
dereklwood: aligned fields with fog vars, and prepended google_ to non-fog vars

provisioner/worker/plugins/providers/fog_provider/fog_provider.json
dereklwood: done

albertshau:
:+1:

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -0,0 +1,319 @@
...
+
+      # do we need sudo bash?
+      sudo = 'sudo' unless @task['config']['ssh-auth']['user'] == 'root'
+
+      set_credentials(@task['config']['ssh-auth'])
+
+      # login with pseudotty and turn off sudo requiretty option
+      log.debug "Attempting to ssh to #{bootstrap_ip} as #{@task['config']['ssh-auth']['user']} with credentials: #{@credentials} and pseudotty"
+      Net::SSH.start(bootstrap_ip, @task['config']['ssh-auth']['user'], @credentials) do |ssh|
+        cmd = %Q[#{sudo} sed -i -e 's/^\\(Defaults\\s\\+requiretty.*\\)$/#\\1/i' /etc/sudoers]
```
wolf31o2: You could do a match + replace versus using backreferences. Using `\s` isn't portable, as it isn't part of the POSIX standard, but that may not be an issue, since we're sure we're on Linux (and not Mac OS). Using the `[[:space:]]` character class is a better match. Also, aren't the options case-sensitive?

``` bash
sed -i -e '/^Defaults[[:space:]]*requiretty/ s/^/#/'
```

wolf31o2:
One minor comment, otherwise... :+1: LGTM

PR #479, opened 2014-09-04
removing 'X-Loom-' from the headers.

albertshau:


albertshau:
ignore, going replace 'X-Loom' with 'Coopr'

gaarf:
IMHO we still need to namespace headers somehow... As @wolf31o2 said, it makes sense to have the product name here.

PR #480, opened 2014-09-04
Update copyright

awholegunch:
Updates copyright statements in all files except docs.


albertshau:
unfortunately, the provisioner directory includes a lot of third party files which also have copyrights that were edited.  We should make sure to only edit ours.

albertshau:
this is probably also true of the ui directory

awholegunch:
Need to try this again. Too many mistakes.

PR #481, opened 2014-09-04
Update loom_service_runner cookbook

wolf31o2:
- check if a service resource exists before attempting to perform an action on it


albertshau:
:+1

PR #482, opened 2014-09-05
add excludes to avoid unpredictability on file copy source

dereklwood:
avoid telling maven to copy two different files to the same location.


albertshau:
:+1:

PR #486, opened 2014-09-05
adding all the fields so they show up in the default provider for the us...

dereklwood:
...er to edit

quick fix so that the standalone default google provider is customizeable.


wolf31o2:
:+1: LGTM...

PR #487, opened 2014-09-05
Update copyrights

awholegunch:
A second attempt; fewer files altered (547 versus 906). Docs are altered in a different PR.


wolf31o2:
Most of those cookbooks have their own repositories and need to be modified, there, first. Otherwise, we will overwrite these changes whenever we make updates to those repositories and import the cookbooks.

albertshau:
@wolf31o2 can you make those changes in the other repos? This change lgtm.

PR #494, opened 2014-09-06
reorder gem statement

dereklwood:
- [x] gem statement needs to be before the require.  otherwise it fails if newer version of fog is also installed


wolf31o2:
LGTM

PR #495, opened 2014-09-06
prevent leaking disks by persisting on first delete attempt

dereklwood:
new delete logic to prevent "leaking" of failed disk deletes on google compute.  
- On first delete attempt, write the currently attached disk names to the task payload (if they don't already exist).
- Then for server and disks, do lookup/delete
- confirms each by expecting wait_for to either detect state change or raise notfound exception.


wolf31o2:
Can we just use `disks` versus `disks_to_delete`?

dereklwood:
was trying to distinguish it from the `disks` that would be used for full-fledged disk support, but i guess at this point it doesn't matter

albertshau:
:+1:

PR #497, opened 2014-09-08
cleaning up some server code. No functional changes.

albertshau:


server/distribution/etc/loom/conf.dist/loom-site.xml.example
```diff
@@ -26,7 +26,7 @@
   <property>
     <name>server.host</name>
     <value>123.456.0.1</value>
-    <description>Specifies the loom host to bind to</description>
+    <description>Specifies the host the server should bind to</description>
```
dereklwood, line 29: "..._address_ the server should bind to" would be more clear

dereklwood:
minor suggestion otherwise :+1: lgtm

PR #498, opened 2014-09-08
launch plugin registration worker via worker_launcher

dereklwood:
- [x] fix to make `--register` launch a worker in the same way that a regular worker is spun up.  This honors any configuration settings, in particular logging settings (which were ignored previously)


albertshau:
:+1:

PR #500, opened 2014-09-09
Coopr NG

gaarf:
Initial dump of coopr-ng.


nitinmotgi:
Merge it and please exclude it from build. Not reviewed it,  but should be fine to merge.

wolf31o2:
Current builds will not do anything with this directory, so it is safe to merge in that regard.

PR #501, opened 2014-09-09
Feature/provider resources

dereklwood:
this PR improves support for uploading of plugin credentials, and passing the necessary info to the plugins.  google provider updates included in this PR, other plugins will follow in separate PRs

changes to provisioner worker:
- [x] sets the current working directory to the plugin's resource work dir, for the duration of the plugin's execution
- [x] bootstrap now relies on `task['config']['automators']`.  removed the default case of running bootstrap for all known automators, since this should never happen, and the code to handle this case was getting ugly.

changes to the google provider:
- [x] class variables to define the known relative paths for resources (matching those defined in the json)
- [x] constructing the resource path from those, and validating up front.

### DIFF
--- a/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
+++ b/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
@@ -25,7 +25,7 @@ class FogProviderGoogle < Provider
 
   # plugin defined resources
   @@p12_key_dir = 'p12_keys'
-  @@ssh_key_dir = 'ssh_private_keys'
+  @@ssh_key_dir = 'ssh_keys'
 
   def create(inputmap)
     @flavor = inputmap['flavor']
@@ -44,7 +44,6 @@ class FogProviderGoogle < Provider
       validate!
       # Create the server
       log.debug "Creating #{@providerid} on Google using flavor: #{flavor}, image: #{image}"
-      log.debug "working dir: #{Dir.pwd}"
 
       # disks are managed separately, so CREATE must first create and confirm the disk to be used
       # handle boot disk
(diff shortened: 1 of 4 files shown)


provisioner/worker/provisioner.rb
```diff
@@ -127,34 +132,35 @@ def delegate_task(task, pluginmanager)
...
+      # server must specify which bootstrap handlers need to run
       log.debug "Task #{task_id} running specified bootstrap handlers: #{task['config']['automators']}"
       task['config']['automators'].each do |automator|
-        classes.push(pluginmanager.getHandlerActionObjectForAutomator(automator))
+        clazz = Object.const_get(pluginmanager.getHandlerActionObjectForAutomator(automator))
+        object = clazz.new(@plugin_env, task)
+        cwd = File.join(@plugin_env[:work_dir], @plugin_env[:tenant], 'automatortypes', automator)
+        Dir.chdir(cwd) do
+          result = object.runTask
+        end
```
albertshau: looks like there is some repeat code that could be pulled out into a common function. Lines 134-138, 141-145, 153-158.

albertshau:
small comment, otherwise lgtm

provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -307,9 +312,15 @@ def validate!
     unless @google_client_email =~ /.*gserviceaccount.com$/
       errors << 'Invalid service account email address. It must be in the gserviceaccount.com domain'
     end
-    [@google_key_location, @google_ssh_keyfile].each do |key|
-      next if File.readable?(key)
-      errors << "cannot read specified key location: #{key}"
+    if @google_ssh_key_name.nil? || @google_p12_key_name.nil?
+      errors << "Fields 'P12 key name' and 'ssh private key name' must be defined'"
```
wolf31o2: This cannot happen, since they're both required, anyway.

dereklwood:
comments addressed... merging

PR #504, opened 2014-09-09
add a status filter to get clusters call.

albertshau:


server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
```diff
@@ -95,7 +98,10 @@ private ClusterHandler(TenantStore tenantStore,
   }
 
   /**
-   * Get all clusters visible to the user.
+   * Get all clusters visible to the user. Clients can include a status filter as an http param. The key is 'status'
+   * and the value is a comma separated list of statuses. Clusters returned must be in one of the statuses given. If
+   * not status param is given, clusters with any status will be returned. Valid values to include in a status filter
```
wolf31o2: s/not/no/

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
```diff
@@ -794,4 +803,21 @@ private JsonObject formatJobPlan(ClusterJob job) throws IOException {
...
+    Set<Cluster.Status> filter = Sets.newHashSet();
+    Map<String, List<String>> queryParams = new QueryStringDecoder(request.getUri()).getParameters();
+    if (queryParams.containsKey("status")) {
+      String statusStr = queryParams.get("status").get(0);
+      String[] statuses = statusStr.split(",");
+      for (String status: statuses) {
+        try {
+          filter.add(Cluster.Status.valueOf(status.toUpperCase()));
+        } catch (IllegalArgumentException e) {
+          LOG.info("Unknown cluster status " + status + " requested.");
```
wolf31o2: warn?

awholegunch: If you use the parameterized version of LOG, you get an optimization when the log level is changed:
LOG.info("Unknown cluster status {} requested.", status);
Plus, the code is often cleaner. This is the standard for CDAP.

albertshau: yup, should be parameterized, forgot it here.  I don't think it should be warn, because its not something concerning for the server itself but an indication of some bad user input.

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
albertshau: done

server/src/main/java/com/continuuity/loom/store/cluster/ClusterStoreView.java
```diff
@@ -23,14 +23,23 @@
...
-   * @throws IOException if there was a problem getting the clusters.
+   * @throws IOException if there was a problem getting the cluster summaries.
    */
   List<ClusterSummary> getAllClusterSummaries() throws IOException;
 
   /**
+   * Get an immutable list of summaries of all clusters in the store that are in one of the given cluster states.
+   *
+   * @param states returned clusters must be in one of these states.
+   * @return Immutable list of summaries of all clusters in the store that are in one of the given cluster states.
```
wolf31o2, line 37: We don't separate them by state in any way, correct? Should we?

albertshau, line 37: correct.  If we have a use case for it we can add it, but for now I don't think it's necessary.

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
wolf31o2: Alright... :-)

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
```diff
@@ -95,7 +98,10 @@ private ClusterHandler(TenantStore tenantStore,
   }
 
   /**
-   * Get all clusters visible to the user.
+   * Get all clusters visible to the user. Clients can include a status filter as an http param. The key is 'status'
+   * and the value is a comma separated list of statuses. Clusters returned must be in one of the statuses given. If
+   * no status param is given, clusters with any status will be returned. Valid values to include in a status filter
```
poornachandra: clusters with any status will be returned -> all clusters will be returned

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
```diff
@@ -794,4 +803,21 @@ private JsonObject formatJobPlan(ClusterJob job) throws IOException {
 
     return jobJson;
   }
+
+  private Set<Cluster.Status> getStatusFilter(HttpRequest request) {
+    Set<Cluster.Status> filter = Sets.newHashSet();
+    Map<String, List<String>> queryParams = new QueryStringDecoder(request.getUri()).getParameters();
+    if (queryParams.containsKey("status")) {
+      String statusStr = queryParams.get("status").get(0);
```
poornachandra, line 812: get(0) can throw exception if status is present but empty. Not sure if it can happen though.

poornachandra:

server/src/main/java/com/continuuity/loom/http/handler/ClusterHandler.java
albertshau, line 812: true, better to check.

PR #507, opened 2014-09-09
quick fix for register require

dereklwood:
fix for bug introduced in #498. i didn't catch it because the correct file gets required transitively also


wolf31o2:
LGTM

PR #508, opened 2014-09-09
Feature/standalone build

dereklwood:
updating standalone build after init scripts moved into their respective component directories
- [x] updated loom.sh to refer to [component]/bin/*.sh
- [x] added missing server/bin to standalone.xml and removed top-level bin pattern which no longer matched anything

final result:

```
./bin/loom.sh
./provisioner/bin/data-uploader.rb
./provisioner/bin/provisioner.sh
./server/bin/dummy-provisioner.sh
./server/bin/server.sh
./ui/bin/ui.sh
```

### DIFF
--- a/standalone/assembly/standalone.xml
+++ b/standalone/assembly/standalone.xml
@@ -37,6 +37,10 @@
             </includes>
         </fileSet>
         <fileSet>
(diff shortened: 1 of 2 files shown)


wolf31o2:
:+1: LGTM

PR #509, opened 2014-09-09
Feature/use embedded ruby

wolf31o2:
Default LOOM_RUBY to the embedded/bin/ruby, but fallback to just ruby if it cannot be found


dereklwood:
:+1:

PR #510, opened 2014-09-09
Feature/resource chef solo updates for cwd, resources

dereklwood:
updates to chef-solo automator to:
- [x] use the resources synced to the (now set) current working directory 
- [x] define its static vars top-level instead of overriding initialize, and other minor cleanup
- [x] fixes #472 - skips any tarball operation for missing resources (logs a warning)


wolf31o2:
:+1: LGTM

PR #513, opened 2014-09-10
Feature/resource joyent

dereklwood:
joyent provider updates:
- [x] change key path to resource name
- [x] latest images
- [x] hack to get centos working again.  if /dev/vdb exists we attempt to remount it, which fails on centos.  the hack checks /proc/mounts for /dev/vdb\* mounted on /data.  open to suggestion on how to make this better

### DIFF
--- a/config/defaults/imagetypes/centos6.json
+++ b/config/defaults/imagetypes/centos6.json
@@ -15,7 +15,7 @@
       "image": "centos-6-v20140718"
     },
     "joyent": {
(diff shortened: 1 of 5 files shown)


wolf31o2:
Can we make the resource name consistent between providers? Also, is that the latest Certified Ubuntu?

Otherwise, LGTM

dereklwood:
yes and yes...

wolf31o2:
:+1: merging...

PR #514, opened 2014-09-10
updating rackspace keyfile to keyfile_name data resource

dereklwood:
rackspace update to use data resource instead of path to keyfile


wolf31o2:
Can we use `ssh_keys` for resource name? Otherwise, LGTM...

wolf31o2:
:+1: merging...

PR #518, opened 2014-09-10
invoking tar with -h which is consistent between gnu and bsd tar

dereklwood:
making the tar shell out work with both gnu tar and bsd tar (mac).  Filed issue #517 to get rid of the shell out.


wolf31o2:
:+1: LGTM... and merging

PR #519, opened 2014-09-10
Remove debug logging of credentials in connection def

wolf31o2:
We do not need to expose credentials at every connection. Removing the debugging output, since it's not necessary and can leak credentials.


albertshau:
:+1:

PR #522, opened 2014-09-10
use explicit logic for deserializing a cluster operation request

albertshau:
so the constructor is actually called and null checks performed.


poornachandra:
LGTM :+1:

PR #527, opened 2014-09-11
Verify that xvda is true, rather than assume it is. Fixes #525

wolf31o2:
Title says it all.


provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/aws.rb
```diff
@@ -196,8 +196,10 @@ def confirm(inputmap)
             ssh_exec!(ssh, "#{sudo} /sbin/mkfs.ext4 /dev/md0 && #{sudo} mkdir -p /data && #{sudo} mount -o _netdev /dev/md0 /data", 'Mounting /dev/md0 as /data')
           elsif xvdb
             ssh_exec!(ssh, "mount | grep ^/dev/xvdb 2>&1 >/dev/null && #{sudo} umount /dev/xvdb && #{sudo} /sbin/mkfs.ext4 /dev/xvdb && #{sudo} mkdir -p /data && #{sudo} mount -o _netdev /dev/xvdb /data", 'Mounting /dev/xvdb as /data')
-          else
+          elif xvda
             ssh_exec!(ssh, "mount | grep ^/dev/xvda2 2>&1 >/dev/null && #{sudo} umount /dev/xvda2 && #{sudo} /sbin/mkfs.ext4 /dev/xvda2 && #{sudo} mkdir -p /data && #{sudo} mount -o _netdev /dev/xvda2 /data", 'Mounting /dev/xvda2 as /data')
+          else
+            log.debug 'No additional instance store disks detected'
```
albertshau, line 202: is everything normal if we get here, or is it a sign that something may be wrong? Just trying to see if debug is the appropriate log level

wolf31o2, line 202: Normal. Could be an EBS-backed instance, which is why Derek filed #525

albertshau:
:+1:

PR #529, opened 2014-09-11
Feature/aws ebs disk

wolf31o2:
Support for EBS-backed instances. Tested with official CentOS 6 AMIs.
- [x] Detect EBS-backed AMI
  - [x] Support setting volume size
  - [x] Support setting `Delete on terminate` option
- [x] Resize EXT3/EXT4 volumes
- [x] Fall-back to instance store


dereklwood:
:+1:

PR #530, opened 2014-09-12
Feature/full path resources

dereklwood:
- [x] when writing `ssh-auth['identityfile]` back to the payload, expand to absolute path
- [x] changing the class variables (`@@ssh_key_dir`) to "class instance variables" with accessor methods.  this is generally recommended as it is more flexible in inheritence
- [x] minor cleanup on aws default provider


albertshau:
:+1:

PR #531, opened 2014-09-12
correct exception class name

dereklwood:
fixing wrong exception class name in a rescue block in the shell provisioner


albertshau:
:100:

PR #535, opened 2014-09-12
support for floating ip

dereklwood:
- [x] associate floating ip if requested and use it as bootstrap ip
- [x] attempt to use private ip for bind_v4 address, default to bootstrap ip


wolf31o2:
LGTM

PR #538, opened 2014-09-12
added CSS minification build step

gaarf:
when running `gulp minify`, CSS will now be minified along with JS.

### DIFF
--- a/ngui/Gulpfile.js
+++ b/ngui/Gulpfile.js
@@ -158,7 +158,7 @@ gulp.task('clean', function() {
 
 
 gulp.task('js:minify', ['js'], function() {
-  return gulp.src('./dist/bundle/*.js')
+  return gulp.src('./dist/bundle/{app,lib}.js')
     .pipe(plug.uglify())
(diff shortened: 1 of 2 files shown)


ngui/Gulpfile.js
```diff
@@ -164,7 +164,14 @@ gulp.task('js:minify', ['js'], function() {
     .pipe(plug.size({showFiles:true, gzip:true, total:false}));
 });
 
-gulp.task('minify', ['js:minify']);
+gulp.task('css:minify', ['css'], function() {
+  return gulp.src('./dist/bundle/*.css')
+    .pipe(plug.minifyCss({keepBreaks:true}))
```
ajainarayanan, line 169: If we are to minify css why keep line breaks?

ajainarayanan:
:+1: LGTM.. Looks like simple change.

ngui/Gulpfile.js
gaarf, line 169: slightly more readable at a small cost... but yeah probably pointless with all the concatenating and post-processing... Eventually we can add sourcemaps.

PR #539, opened 2014-09-12
updating username to user

dereklwood:
default provider used `ssh_username` instead of `ssh_user` used everywhere else


wolf31o2:
:+1: LGTM

PR #540, opened 2014-09-12
adding google-api-client gem with same requirements as fog

dereklwood:
updating the fog provider plugin gemfile with google-api-client, same versions as fog.


wolf31o2:
:+1: LGTM

PR #544, opened 2014-09-12
first review of ngui

rns4731:
First review of ngui.
- Standardizes use of https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml wherever possible with comments, spacing, naming conventions etc.
- Changes .tpl files to .html and corresponding build targets.
- Contains a bunch of TODOs, this PR should only be merged once there are no [TODO]s remaining in these files.


gaarf:
I'm not sure about moving the constants out into a seperate file, everything else looks good to me.

gaarf:
Oh also if you rename the directive templates from `.tpl` to `.html`, then the Gulpfile needs to be modified accordingly. I went with the `.tpl` extension because it maps nicely to the generated `tpl.js` file and the `tpl` gulp task.

rns4731:
@gaarf constants are injected everywhere but it wasn't intuitive where to find them without just grepping on the project. I don't think it makes sense to define them wherever they're first used. I changed the .tpl to .html and in the gulpfile, everything compiles and runs now, let me know if it doesn't for you.

There are a few todos here that need to be completed and I can't finish all of them. Please take a look and fill accordingly or state why not. At a minimum every file and unintuitive method needs to have a header/comment.

gaarf:
constants are not defined where they are first used, they are defined in same file as service they originate from, and are named accordingly...

If we move them out, won't that hinder our ability to easily share services across projects?

rns4731:
@gaarf @ajainarayanan added first directive: jsoneditor to be pulled from shared repo https://github.com/continuuity/spikes/tree/develop/frontend-shared-components
Should not change the functionality of the app, only modified a few settings.

gaarf:
@nistala I can't access that repo :-(

rns4731:
The repo I've stated is temporary because we haven't gotten approval yet to set up a shared frontend repo. But I am assuming that shared repo will be private.

wolf31o2:
The shared repo cannot be private. For this to be merged, all functionality must be accessible to external entities. We do our tests, for one, using public infrastructure. Just letting you know that this will be a requirement of the shared frontend repo, if we're using it, directly, in other projects.

gaarf:
This PR was for discussion only. Most of @nistala recommendations will be implemented in an different upcoming branch. Closing this.

PR #545, opened 2014-09-12
fail with nicer error in case of google quota exceeded

dereklwood:
fail with nicer error in case of nil server, and suggest quota as likely reason.


wolf31o2:
:+1: LGTM... merging

PR #547, opened 2014-09-12
Adrien's NGUI work in progress

gaarf:
Since I'm working on several items at the same time on NGUI, this PR can serve as a single place to review ongoing changes.

### DIFF
--- a/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
+++ b/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
@@ -1,5 +1,5 @@
 angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl', 
-  function ($scope, CrudFormBase, $state, $modal, myApi, $timeout) {
+  function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, myApi, $timeout) {
     CrudFormBase.apply($scope);
 
 
@@ -17,51 +17,6 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
     else {
       failure();
     }
-
-
-
-
-    function failure () {
-      $state.go('404');
-    }
-
-
-
-
-    function update () {
-      $scope.model.$get();
-    }
-
-
-
-
-
-    function doActionsModal (nodeId) {
-      $state.go('clusters.detail.node', {nodeId: nodeId});
-
-      var modalScope = $scope.$new(true);
-
-      modalScope.node = $scope.model.nodes.filter( function(node) {
-        return node.id === nodeId;
-      })[0];
-
-      console.log(modalScope.node.actions);
-
-      modalScope.$on('modal.hide', function () {
-        $state.go('^');
-      });
-
-      $modal({
-        title: nodeId,
-        contentTemplate: '/partials/clusters/detail-node.html', 
-        placement: 'center',
-        scope: modalScope,
-        show: true
-      });
-    }
-
-    $scope.doActionsModal = doActionsModal;
-
 
 
 
@@ -107,9 +62,55 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
     });
 
 
+    $scope.$on(MYSERVICEPICKER_EVENT.manage, update);
+
+
+
     $scope.doSubmitServices = function (arrSvcs) {
       myApi.ClusterService.save( {clusterId: $scope.model.id}, { services: arrSvcs }, update);
     };
+
+
+
+    $scope.doActionsModal = doActionsModal;
+
+
+
+    function doActionsModal (nodeId) {
+      $state.go('clusters.detail.node', {nodeId: nodeId});
+
+      var modalScope = $scope.$new(true);
+
+      modalScope.node = $scope.model.nodes.filter( function(node) {
+        return node.id === nodeId;
+      })[0];
+
+      console.log(modalScope.node.actions);
+
+      modalScope.$on('modal.hide', function () {
+        $state.go('^');
+      });
+
+      $modal({
+        title: nodeId,
+        contentTemplate: '/partials/clusters/detail-node.html', 
+        placement: 'center',
+        scope: modalScope,
+        show: true
+      });
+    }
+
+
+    function failure () {
+      $state.go('404');
+    }
+
+
+
+
+    function update () {
+      $scope.model.$get();
+    }
 
   }
 );
--- a/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
+++ b/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
@@ -90,7 +90,7 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
         $scope.leaseExtendMs = 0;
         update();
       });
-    }
+    };
 
 
 
@@ -116,8 +116,6 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
       modalScope.node = $scope.model.nodes.filter( function(node) {
         return node.id === nodeId;
       })[0];
-
-      console.log(modalScope.node.actions);
 
       modalScope.$on('modal.hide', function () {
         $state.go('^');
--- a/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
+++ b/ngui/app/js/controllers/crud/cluster-detail.ctrl.js
@@ -69,9 +69,11 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
     $scope.$watchCollection('model', function (data) {
       if(!data.$resolved) { return; }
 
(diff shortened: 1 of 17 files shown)


gaarf:
implements the "node actions" modal within cluster details' service-sets section.

screen shot 2014-09-12 at 4 38 55 pm

gaarf:
implements "Click to show password"

screen shot 2014-09-13 at 6 15 48 pm

gaarf:
UI for lease extension

expiryanim

gaarf:
@ajainarayanan please review :-)

ngui/app/js/directives/servicepicker/servicepicker.js
```diff
@@ -26,7 +30,7 @@ module.directive('myServicePicker', function myServicePickerDirective () {
 
       $scope.manageService = function (action, name) {
         myApi.ClusterService[action]( {clusterId: $scope.clusterId}, {name: name}, function () {
-          console.log('manageService', action, name, arguments); // FIXME - broadcast?
+          $scope.$parent.$emit(MYSERVICEPICKER_EVENT.manage, action, name);
```
gaarf, line 33: TBD: should this be emitted on the parent, on $rootScope, or something else ?

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
```diff
@@ -1,33 +1,42 @@
 angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl', 
-  function ($scope, CrudFormBase, $state, myApi, $timeout) {
+  function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, myApi, $timeout, moment) {
```
ajainarayanan, line 2: I am guessing `MYSERVICEPICKER_EVENT` is not a constant. Any reason why it is all caps?

ajainarayanan, line 2: I also assumed that in angular world you add `$` to any variable being passed to understand that it is being injected by angular. 

Am I understanding it wrong because some variables here are preceded by $ and some are not. Are we following some best practices on it that I am failing to see?

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
```diff
@@ -52,16 +61,90 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
...
+    $scope.leaseExtendMs = 0;
+
+    $scope.$watch('leaseExtendMs', function (ms) {
+      ms = parseInt(ms, 10); // work around input[type=range] ng-model being a string
+      $scope.leaseExtendDate = moment($scope.model.expireTime).add(ms, 'ms').toDate();
+      $scope.leaseExtendHumanized = moment.duration(ms, 'ms').humanize();
     });
 
+    $scope.doLeaseExtend = function () {      
+      myApi.Cluster.save({id: $scope.model.id}, {expireTime: $scope.leaseExtendDate.valueOf() }, function () {
```
ajainarayanan: It would be nicer to avoid single line js objects. 

``` javascript
myApi.Cluster.save({
    id: $scope.model.id
}, {
    expireTime: $scope.leaseExtendDate.valueOf()
}, function() {
  ...
});
```

Something like this. Makes a little bit readable at a first glance.

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
```diff
@@ -52,16 +61,90 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
...
+        $alert({title:'Lease extended until:', content: moment($scope.leaseExtendDate).format('LLL'), type:'success', duration:3});
```
ajainarayanan: Same here too.

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
gaarf, line 2: `MYSERVICEPICKER_EVENT` absolutely is a constant, that's why it is all caps.

---

Re: the $ prefix, it is supposedly only for "angular internal" injectables. so like `$timeout` and `$scope`, those are angular objects made by angular team, they are supposed to have the `$`.

`$state` and `$alert` are from libraries we use, their authors have chosen to prefix their module with `$` because they want to be angular-y. Strictly speaking I think they should NOT have the prefix but that's the way the authors chose to do it.

Our internally-grown injectables should NOT use the prefix.

hth?

ngui/app/js/directives/focus/focus.js
```diff
@@ -7,15 +7,23 @@ module.directive('myFocus', function myFocusDirective ($timeout, myFocusManager)
...
-          $timeout(function() {
-            element[0][o.focus?'focus':'select']();
-          });
-        }
+      attrs.$observe('myFocus', function (newVal) {
+        // console.log('[myFocus] watching', newVal);
+
+        var cleanup = myFocusManager.is.$watch(newVal, function (o) {
+          if(o && (o.focus || o.select)) {
+            $timeout(function() {
```
ajainarayanan, line 15: I am guessing from the documentation the default (if no argument is passed for delay) is 0. If that is the case Is there any specific reason why this function is just wrapped in $timeout?

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
```diff
@@ -52,16 +61,90 @@ angular.module(PKG.name+'.controllers').controller('ClusterDetailCtrl',
...
+
+
+
     $scope.doSubmitServices = function (arrSvcs) {
       myApi.ClusterService.save( {clusterId: $scope.model.id}, { services: arrSvcs }, update);
     };
 
+
+
+    $scope.doActionsModal = doActionsModal;
```
ajainarayanan, line 121: Any reason why the assignment and the actual function are separated. Causes no harm I guess but just curious.

ngui/app/js/directives/focus/focus.js
gaarf, line 15: yes, focus() needs to be called asynchronously. The reason is that a single controller may call MyFocusManager.focus() several times, we will always want the last one called to have precedence.

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
gaarf, line 121: `doActionsModal()` is called directly at the top of the file, line#12

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
ajainarayanan, line 2: Ok that should be fine. But I am not sure about the name of the variable. It suggests it is an event to me for an instant but it is actually a constant. I am guessing from the 'My' terminology it is a service? Or is it a event name? Not clear to me in one glance. I think you have much more better context and if it makes sense then it makes sense.

ngui/app/js/directives/focus/focus.js
```diff
@@ -7,15 +7,23 @@ module.directive('myFocus', function myFocusDirective ($timeout, myFocusManager)
...
+              element[0][o.focus?'focus':'select']();
```
ajainarayanan: Two concerns. 
- Whenever the newVal is updated we are adding a a function to $timeout but I am not sure where we are resolving the promise that is being returned from $timeout. So if my guess is right if the newVal changes 4 times we are adding/creating 4 timeout promises and they are executed consecutively?
- `element[0][o.focus? 'focus': 'select']()`  Not sure if it is a little less intuitive but at the end I am not sure if I understand why we need to do this`element[0]['focus']()`  instead of 

``` javascript
if (o.focus === "select") {
  element[0].focus();
}  else if (o.focus === "select") {
  element[0].select();
}
```

Not a huge fan of ternary operator. Chances of screw-ups are a little high while minification.

ajainarayanan: I don't understand they wrap the element in jq-lite object and we still extract it out and use the native method of the element :disappointed:

ajainarayanan:
Minor comments but it is good to be merged, :+1: Two thumbs up for the gif preview. :)

ngui/app/js/controllers/crud/cluster-detail.ctrl.js
gaarf, line 2: it's actually a hashmap of event names... so it's used like this: `MYSERVICEPICKER_EVENT.manage`, and that will return a string. Agreed it's a little akward!

ngui/app/js/directives/focus/focus.js
gaarf: I'll remove the ternary operator, you're right it's not readable.

---

Re: jqLite, the thing is that it doesnt expose a `.focus()` nor a `.select()`... it has a `.triggerHandler()` but that doesn't do what we want (the native browser behavior).

ajainarayanan: Talked to @gaarf offline. :+1: Looks good.

PR #548, opened 2014-09-13
Always exit 0, even if last if does not match

wolf31o2:
Otherwise, the final test doesn't match, and we exit with 1 (false) because we had no failures to match. Oops!


dereklwood:
:+1: oops indeed

PR #549, opened 2014-09-13
update the mongodb template to use 'mongodb' as the service all

albertshau:
the time since its 'mongodb' on ubuntu but 'mongod' on rhel.


wolf31o2:
:+1: LGTM...

PR #555, opened 2014-09-15
as workaround until validation.pem is a resource, look for it in current...

dereklwood:
... working dir, which is scripts


wolf31o2:
:+1: LGTM

PR #556, opened 2014-09-15
hbase services may run into install errors if zookeeper-server

albertshau:
is not installed first. This is a workaround, as
ideally the cookbook/service would install without an ordering
requirement.


wolf31o2:
:+1: LGTM

PR #560, opened 2014-09-16
Feature/top level tests

wolf31o2:
This is just the framework. It needs to be updated with actual tests. Currently, `rubocop` will fail due to presence of `.rubocop.yml` and `.rubocop_todo.yml` files in subdirectories.


provisioner/.rubocop_todo.yml
```diff
@@ -0,0 +1,277 @@
+# This configuration was generated by `rubocop --auto-gen-config`
+# on 2014-09-15 19:15:16 -0700 using RuboCop version 0.24.1.
```
albertshau, line 2: does this need to be checked in?

wolf31o2, line 2: Need? No. The version is helpful, because it tells you what we used, but this isn't necessary. That being said, the file is autogenerated, as it says in the header.

albertshau, line 2: I think we should remove it, seems similar to checkstyle results the java parts generate.

albertshau:
1 comment, otherwise lgtm

provisioner/.rubocop_todo.yml
wolf31o2, line 2: So, autogenerate, then edit? This file is actually data for rubocop, not something like test results... (though it looks like it)

PR #561, opened 2014-09-16
Remove old compatibility code for Loom 0.9.7 and below

wolf31o2:
The fog_provider plugin now requires 0.9.8 or better for full functionality. Due to this, removing compatibility for older releases from the plugin.


dereklwood:
:+1:

PR #562, opened 2014-09-16
Feature/ngui env config

gaarf:
fixes #554

adds a way to send runtime config to the client + use https://github.com/expressjs/morgan for logging.

### DIFF
--- a/ngui/app/js/services/api.js
+++ b/ngui/app/js/services/api.js
@@ -1,10 +1,21 @@
 var module = angular.module(PKG.name+'.services');
 
 module.factory('MYAPI_PREFIX', function($location, MY_CONFIG){
-  return $location.protocol() + '://' +
-            $location.host() + ':' + MY_CONFIG.COOPR_CORS_PORT +
-            MY_CONFIG.COOPR_SERVER_URI.replace(/^(https?:\/\/)/i, '/') + 
-            '/v2/';
+
+  // to work with CORS proxy, we expect that the URI will include a port
+  //  ... first we need to remove the protocol from the URI 
+  var restPath = MY_CONFIG.COOPR_SERVER_URI.replace(/^(https?:\/\/)?/i, '');
+  if(restPath.substr(-1)!=='/') {
+    restPath += '/'; // then we ensure it ends with a slash
+  }
+
+  /* end result should look something like:
+
+       http :// host.foo.com : 8081 / otherhost.foo.com:55054/v2/
+  */
+  return  $location.protocol() + '://' + $location.host() + 
+            ':' + MY_CONFIG.COOPR_CORS_PORT + '/' + restPath;
+            
 });
 
 
--- a/ngui/app/js/services/api.js
+++ b/ngui/app/js/services/api.js
@@ -1,9 +1,10 @@
 var module = angular.module(PKG.name+'.services');
 
 module.factory('MYAPI_PREFIX', function($location, MY_CONFIG){
-  console.log(MY_CONFIG);
-  return $location.protocol() + '://' + $location.host() + 
-                ':8081/0.0.0.0:55054/v2/';
+  return $location.protocol() + '://' 
+            + $location.host() + ':' + MY_CONFIG.COOPR_CORS_PORT
+            + MY_CONFIG.COOPR_SERVER_URI.replace(/^(https?:\/\/)/i, '/') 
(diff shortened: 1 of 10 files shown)


ngui/app/js/services/api.js
```diff
@@ -1,8 +1,10 @@
 var module = angular.module(PKG.name+'.services');
 
-module.factory('MYAPI_PREFIX', function($location){
-  return $location.protocol() + '://' + $location.host() + 
-                ':8081/0.0.0.0:55054/v2/';
+module.factory('MYAPI_PREFIX', function($location, MY_CONFIG){
+  return $location.protocol() + '://' +
```
ajainarayanan: Can we have a sample url that will generated by this expression in comments, something like what each expression might evaluate to in a more common scenario (like localhost)

ngui/server.js
```diff
@@ -1,37 +1,70 @@
...
+      COOPR_CORS_PORT: COOPR_CORS_PORT,
+      COOPR_UI_PORT: COOPR_UI_PORT
+
+    });
+
+
+var hilite = function (v) { return '\x1B[7m' + v + '\x1B[27m'; },
     httpLabel = '\x1B[40m\x1B[32mhttp-server\x1B[39m\x1B[49m',
-    corsLabel = '\x1B[40m\x1B[35mcors-anywhere\x1B[39m\x1B[49m';
+    corsLabel = '\x1B[40m\x1B[35mcors-proxy\x1B[39m\x1B[49m',
```
ajainarayanan: I understand this is for logs. Can we have it a little less intimidating? Something like this,

``` javascript
var formatting = {
     green: '\x1B[40m\x1B[35',
     greenend : '\x1B[39m\x1B[49m',
     othercolors: ' '
};
...
..
corsLabel = formatting.green + 'cors-anywhere' + formatting.greenend;
```

I feel this is a tad easier to look at.

ngui/server.js
```diff
@@ -1,37 +1,70 @@
...
+      function (req, res) {
+        if(req.url !== '/config.js') {
+          // all other paths are passed to ecstatic
+          return res.emit('next');
+        }
+        res.writeHead(200, { 
+          'Content-Type': 'application/javascript',
+          'Cache-Control': 'no-store, must-revalidate'
+        });
+        res.end('angular.module("' + pkg.name + '.config", [])'
```
ajainarayanan, line 51: Does this send js code as response which is directly being executed in the browser?  Can't we send it as two separate files like config.js or and config-constants.js ? Am I missing something here?

gaarf, line 51: yes it is sending down a .js file to be executed in the browser. the file depends on angular.

I'm not sure I follow why you'd want two files ? It's one file that defines one constant. The constant is an hashmap of strings and is named MY_CONFIG.

ngui/app/js/services/api.js
gaarf: yes, good call.

ngui/server.js
```diff
@@ -1,37 +1,70 @@
...
-      '\n%s %s\n\x1B[1m%s\x1B[22m \x1B[36m%s\x1B[39m',
-      label,
-      (new Date).toUTCString(),
-      req.method,
-      req.url
-    );
-  }
-}
-
+  .listen(COOPR_CORS_PORT, '0.0.0.0', function() {
```
wolf31o2, line 72: Does the user directly connect to the CORS proxy port? If so, this will definitely need to be documented as a port that needs to be open for the UI.

gaarf, line 72: yes, the browser connects to the CORS proxy port on every API request.

ngui/server.js
gaarf: ok

ngui/server.js
wolf31o2, line 72: So, to setup high-availability, I would need to do both COOPR_UI_PORT and COOPR_CORS_PORT... Is there any kind of state or other issue that could arise in this scenario?

ajainarayanan:
:+1: JS looks good to be merged.

ngui/server.js
gaarf, line 72: @wolf31o2 what do you mean by _do_ ? 

I've removed COOPR_UI_PORT from the generated /config.js because the client can figure that out easily by looking at window.location ... but it remains configurable on server side.

PR #563, opened 2014-09-16
Feature/epel

wolf31o2:
Install EPEL repository on RHEL platforms. Can be overridden with `base['use_epel']` set to `false`... Fixes #542

### DIFF
--- a/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_base/recipes/default.rb
+++ b/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_base/recipes/default.rb
@@ -20,12 +20,14 @@
 # This forces an apt-get update on Ubuntu/Debian
 case node['platform_family']
 when 'debian'
-  include_recipe "apt::default"
+  include_recipe 'apt::default'
+when 'rhel'
+  include_recipe 'yum-epel::default' if node['base']['use_epel'].to_s == 'true'
 end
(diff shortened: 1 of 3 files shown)


provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_base/recipes/default.rb
```diff
@@ -20,12 +20,14 @@
 # This forces an apt-get update on Ubuntu/Debian
 case node['platform_family']
 when 'debian'
-  include_recipe "apt::default"
+  include_recipe 'apt::default'
+when 'rhel'
+  include_recipe 'yum-epel::default' if node['base']['use_epel'].to_s == 'true'
```
albertshau, line 25: is it normal in cookbooks to use strings for boolean checks instead of trying to convert to a boolean?  For example somebody could have a typo 'treu'.

albertshau:
small question, but lgtm

provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_base/recipes/default.rb
wolf31o2, line 25: Ruby doesn't have a built-in "to boolean" so you have to reimplement one everywhere you want one. I find it easier to force it to a string, which is built-in...

PR #566, opened 2014-09-16
Feature/google create edge cases

dereklwood:
fixes #516 
- [x] persist providerid to result immediately to ensure its there for any delete request
- [x] attempt to use existing disks of identical name/specs (primarily for retry use-case)
- [x] protect against bad values of data_disk_size_gb

### DIFF
--- a/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
+++ b/provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
@@ -326,10 +326,10 @@ class FogProviderGoogle < Provider
       existing_zone_name = disk.zone_name.nil? ? nil : disk.zone_name.split('/').last
       existing_source_image = disk.source_image.nil? ? nil : disk.source_image.split('/').last
       if size_gb == existing_size_gb && zone_name == existing_zone_name && source_image == existing_source_image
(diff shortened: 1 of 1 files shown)


provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/google.rb
```diff
@@ -56,7 +59,7 @@ def create(inputmap)
       @disks = [disk]
 
       # handle additional data disk
-      if fields['google_data_disk_size_gb']
+      if fields['google_data_disk_size_gb'] && fields['google_data_disk_size'].to_i > 0
         data_disk_name = "#{@providerid}-data"
         log.debug "Creating data disk: #{data_disk_name} of size #{fields['google_data_disk_size_gb']}"
         create_disk(data_disk_name, fields['google_data_disk_size_gb'], @zone_name, nil)
```
wolf31o2: I know that this isn't part of your changes, but it showed up in the diff, so....

Either `fields['google_data_disk_size_gb'].to_i` before passing it, so we're sure that create_disk gets an integer, or convert to integer within create_disk.

wolf31o2:
One minor comment, otherwise :+1: LGTM

PR #567, opened 2014-09-17
Feature/ngui auth header

gaarf:
follow-up to #562 re #554
- renames MYAPI_PREFIX to myApiPrefix, because although we use it as a constant, it is a factory and hence not injectable in config blocks.
- delays app bootstrap until we fetch /config.json, which contains runtime config + gives us the authorization header to use on API calls.

@wolf31o2 I believe this implements what we discussed today over hipchat regarding sharing the reverse-proxy auth with CORS port.

### DIFF
--- a/ngui/app/js/services/api.js
+++ b/ngui/app/js/services/api.js
@@ -50,7 +50,7 @@ module.factory('myApi', function(
 
 });
 
-module.config(function ($httpProvider, MYAPI_EVENT, MY_CONFIG) {
+module.config(function ($httpProvider, MYAPI_EVENT) {
   $httpProvider.interceptors.push(function ($q, $timeout, $rootScope, $log, myAuth, myApiPrefix) {
     var isApi = function(url) {
       return url.indexOf(myApiPrefix) === 0;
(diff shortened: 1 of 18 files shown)


gaarf:
\+ update to [latest stable angular.js](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#1225-hypnotic-gesticulation-2014-09-16)

ngui/app/js/services/api.js
```diff
@@ -1,6 +1,6 @@
 var module = angular.module(PKG.name+'.services');
 
-module.factory('MYAPI_PREFIX', function($location, MY_CONFIG){
+module.factory('myApiPrefix', function ($location, MY_CONFIG) {
```
ajainarayanan, line 3: https://docs.angularjs.org/guide/providers#providers_provider-recipe

<serviceId>Factory for naming. That way we know that it is using a factory recipe. Any thoughts? Needn't go in this PR but for later reference.

ajainarayanan:
:+1: Looks good. Good be merged.

ngui/app/js/services/api.js
gaarf, line 3: will name the currently anonymous function `myApiPrefixFactory` in another PR.

PR #571, opened 2014-09-17
include config/[hostname/ipaddresses] in merged chef json

dereklwood:
when chef-solo automator transforms task json into chef json, it will now populate `{"loom": { "hostname": "foo" }, "ipaddresses": { } }`


albertshau:
:100:

PR #572, opened 2014-09-18
Add DNS support to loom_base

wolf31o2:
Currently, we only support dnsimple, but any DNS provider which can be modified via API can be added. By default, this does nothing, so it's safe to add everywhere. To use this, a user would need to add JSON to his `base` service in Loom such as:

``` json
{"loom_dns":{"provider":"dnsimple"}}
```

This would require the use of a `creds` data bag with a `dnsimple` item. These are configurable, and someone can choose to provide an username/password combo, instead.

``` json
{"loom_dns":{"provider":"dnsimple","dnsimple":{"use_databag":false}},"dnsimple":{"username":USERNAME,"password":PASSWORD}}
```

### DIFF
--- a/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/metadata.rb
+++ b/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/metadata.rb
@@ -1,6 +1,6 @@
 name             'loom_dns'
-maintainer       'Cask'
-maintainer_email 'ops@cask.co'
+maintainer       'Continuuity'
+maintainer_email 'ops@continuuity.com'
 license          'All rights reserved'
 description      'Installs/Configures DNS'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
--- a/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/metadata.rb
+++ b/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/metadata.rb
@@ -1,0 +1,10 @@
+name             'loom_dns'
+maintainer       'Cask'
+maintainer_email 'ops@cask.co'
+license          'All rights reserved'
+description      'Installs/Configures DNS'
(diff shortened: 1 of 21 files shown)


provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/metadata.rb
```diff
@@ -0,0 +1,10 @@
+name             'loom_dns'
+maintainer       'Cask'
```
dereklwood: fyi... doesn't match the copyright headers

wolf31o2: Fixed.

provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/recipes/dnsimple.rb
```diff
@@ -0,0 +1,55 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+%w(build-essential dnsimple).each do |recipe|
+  include_recipe recipe
+end
+
+if node['loom_dns']['dnsimple']['use_databag']
```
dereklwood: the `use_databag` attribute seems unecessary and burdensome to have to set to false if you want to use the dnsimple username/password.  why dont we just use `node['loom_dns']['dnsimple']...` if defined, else default to `node['dnsimple']...`, else fail ?

dereklwood:
minor comment otherwise :+1: lgtm

provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/recipes/dnsimple.rb
wolf31o2: Why `node['loom_dns']['dnsimple']` and not just a check directly on `node['dnsimple']`?

PR #573, opened 2014-09-18
ngui gargantuan PR

gaarf:
mostly renaming, adding comments, moving things around as per conversations with @nistala and @ajainarayanan ( cf #544 )

also includes Tenant form feature.


ngui/app/js/controllers/cluster/detail-ctrl.js
```diff
@@ -0,0 +1,163 @@
...
+
+  if($state.params.id) {
+    $scope.model = myApi.Cluster.get( {id:$state.params.id});
+    $scope.model.$promise
+      .then(function (model) {
+        if($state.is('clusters.detail.node')) {
+          doActionsModal($state.params.nodeId);
+        }
+      })
+      ['catch'](failure);
```
ajainarayanan, line 19: `then` returns an object that has a `catch` function for which we can pass `failure`? Can we write a little bit more easy-to-read?

Does `$promise` in angular has a syntax like regular promise something like,

``` javascript
$promise
.then(function() {
...
})
.fail(function(exception) {
...
})
.fin(function (final) {
...
});
```

Or something like this mentioned in the docs https://docs.angularjs.org/api/ng/service/$q

gaarf, line 19: `then()` returns a promise, which has a `then()`, which returns a promise, etc...

if you don't like "catch", we can do `then(function() {...}, failure)`, but when you do that it only works for that point in the promise chain.

gaarf, line 19: `promise.catch(errorCallback)` – shorthand for `promise.then(null, errorCallback)`

ajainarayanan:
:+1: Looks good to me.

PR #576, opened 2014-09-18
get a specific automatortype while waiting since the entire

albertshau:
collection always returns a 200.


wolf31o2:
I wondered about that... :100:

PR #580, opened 2014-09-19
Feature/ui migrate loom

rns4731:
Does a few things:
- Add provider create and edit page, works mostly except I couldn't get ng-required and $valid/$invalid to work on the form.
- pulled out providerfields directive and replaced it with something more generic


ngui/app/js/controllers/cluster/form-ctrl.js
```diff
@@ -17,17 +17,21 @@ function ($scope, $state, $q, myApi, myFocusManager, myHelpers) {
       allImages = myApi.ImageType.query(),
       allServices = myApi.Service.query();
 
+
+  $scope.availableProviderTypes = myApi.ProviderType.query();
```
gaarf: I don't think `availableProviderTypes` needs to be on $scope

ngui/app/js/controllers/cluster/form-ctrl.js
```diff
@@ -74,6 +78,12 @@ function ($scope, $state, $q, myApi, myFocusManager, myHelpers) {
       $scope.chosenProvider = $scope.availableProviders.filter(function (p) {
         return p.name === name;
       })[0];
+
+      $scope.availableProviderTypes.forEach(function (type) {
```
gaarf: why not use Array.filter like above?

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * jsontoformfields directive.
+ *
+ * This takes json of this format and converts them into form fields bound to a provided object.
+ * !! This directive modifies bindobject passed to it from the parent.
+ *
+ * Expected json:
```
gaarf: it doesnt expect JSON, it expects an object.

please don't use "json" in names unless you're doing `JSON.stringify()` or `angular.toJson()`, which is not the case here.

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,98 @@
...
+ * <div my-jsontoformfields fieldsjson="json" bindobject="model.fields"/>
+ */
+
+angular.module(PKG.name+'.directives').directive('myJsontoformfields', 
+function myJsontoformfieldsDirective () {
+  return {
+    restrict: 'AE',
+    replace: true,
+    scope: {
+      fieldsjson: '=',
```
gaarf: call this just "fields" ?

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,98 @@
...
+      bindobject: '='
```
gaarf: to be consistent with other directives, this should be called `model`

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,98 @@
...
+
+      scope.$watch('fieldsjson', function (newVal, oldVal) {
+
+        if (!angular.equals(newVal, oldVal)) {
+          if (scope.fieldsjson && typeof scope.fieldsjson !== 'object') {
+            scope.fieldsjson = angular.fromJson(scope.fieldsjson);
+          }
+
+          if (!scope.fieldsjson.hasOwnProperty('fields')) {
+            throw Error('fields not declared in json for fieldsjson');
```
gaarf: we didn't talk about throwing Errors.... I think directives should never throw, but I guess it is debatable.

ngui/app/js/directives/providerfields/providerfields.js
```diff
@@ -1,80 +0,0 @@
...
-      function setFields () {
-        var fields = {},
-            model = {},
-            providerType = allTypes.filter(function (type) {
-              return type.name === $scope.provider.providertype;
-            })[0];
-
-        typeReqs = [];
-
-        angular.forEach(['user', 'admin'], function (role) {
```
gaarf, line 33: where did this logic move to ?

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
rns4731: line 41, it takes either for convenience

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
rns4731: change to fieldsconfig, fields here could stand for the html fields in the form and it would become fields.fields...

ngui/app/js/directives/providerfields/providerfields.js
rns4731, line 33: not required anymore, this is a more generic directive that accepts the actual fields spec

rns4731:
Just pushed changed for comments, few things still not working:
- validation with $valid and ng-required
- override fields

rns4731:
Fixed:
- validation
- overrides

ngui/app/js/directives/providerfields/providerfields.js
gaarf, line 33: @nistala I insist, please don't change clusters. As is, you've changed the layout of provider fields in clusters page, and maybe I'm wrong but it appears there is now some missing logic.  I want to be able to review _just_ your additions needed for the feature at hand, not have to check if a bug was introduced elsewhere.

_We should not ever delete or change the API of an existing directive_, what if another developer is using said directive in his/her parallel branch?

gaarf:
Please do not change clusters. Add whatever directive you feel is needed, but not change the API or existence of directives that are already there.

Also,  for consistency with rest of UI,  let's not use bs-select, Use `<select>` with ng-options. If we really need icons in the options, we can use bs-dropdown which is way more flexible.

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * jsontoformfields directive.
```
gaarf: still disturbed that is named "json".

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
```diff
@@ -0,0 +1,110 @@
...
+    templateUrl: 'jsontoformfields/jsontoformfields.html',
+    link: function (scope, element, attrs) {
+
+      scope.$watch('fieldsconfig', function (newVal, oldVal) {
+
+        if (!angular.equals(newVal, oldVal)) {
+          if (scope.fieldsconfig) {
+
+            if (typeof scope.fieldsconfig !== 'object') {
+              scope.fieldsconfig = angular.fromJson(scope.fieldsconfig);
```
gaarf: so if it's _not_ JSON but instead is a resource ( the common case), `angular.fromJson` will remove the methods on the prototype of `fieldsconfig`, along with any $-prefixed properties. And since it is 2-way bound, this will affect the scope that consumes this directive. I'm pretty sure we don't want that.

ngui/app/js/services/api/providertypes.js
```diff
@@ -0,0 +1,11 @@
+angular.module(PKG.name+'.services').factory('myApi_providerTypes', 
+function ($resource, myApiPrefix) {
+
+  return {
+
+    ProviderType: $resource(myApiPrefix + 'plugins/providertypes/:name', {})
```
gaarf, line 6: does it really have to be in a different file? It still relates to providers.

ngui/app/js/directives/providerfields/providerfields.js
rns4731, line 33: we can't just say don't modify old code, here it makes sense to do it, there is a very specific directive and it is replaced with a more general directive that can be used in a variety of cases. I'd rather write tests to ensure we don't break stuff but saying don't modify existing code or directives isn't realistic or sustainable, we're just going to end up with a bunch of legacy code and repeats. What is the rationale for not modifying here other than we don't have tests to ensure it works correctly?

gaarf, line 33: we don't have tests to ensure it works correctly ;-)

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
rns4731: angular.fromJson just does a json parse - https://github.com/angular/angular.js/blob/master/src/Angular.js#L1025

ngui/app/js/services/api/providertypes.js
rns4731, line 6: following convention, can change.

ngui/app/js/directives/jsontoformfields/jsontoformfields.js
gaarf: I stand corrected! confused with `toJson`

rns4731:
removed bs-select, renamed directives.
Remaining:
unit test for directives and for controller.

gaarf:
since this doesnt merge after rename, I'm gonna pull out at least the directive contribution in a different branch. Then we can extract @nistala 's work in new PRs.

gaarf:
cf #609

PR #582, opened 2014-09-19
fix not_if statement in dns recipe.

albertshau:


dereklwood:
:+1: for the change, and :+1: for albert doing chef development :)

PR #583, opened 2014-09-19
run apt-update at compile time in case there are other recipes

albertshau:
like the dnsimple recipe that will try to get packages at compile
time.


dereklwood:
:+1:

PR #584, opened 2014-09-20
add ability to whitelist the subdomains that should be added to

albertshau:
dns.

### DIFF
--- a/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/recipes/dnsimple.rb
+++ b/provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/recipes/dnsimple.rb
@@ -49,7 +49,7 @@ end
 # Only register whitelisted subdomains if they are set
 subdomain_whitelist = node['loom_dns']['subdomain_whitelist']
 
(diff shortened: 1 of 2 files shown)


provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/loom_dns/recipes/dnsimple.rb
```diff
@@ -46,13 +46,18 @@
...
-  type 'A'
-  username dnsimple['username']
-  password dnsimple['password']
-  domain subdomain
-  ttl node['loom_dns']['default_ttl']
-  not_if { subdomain == 'local' }
+# Only register whitelisted subdomains if they are set
+subdomain_whitelist = node['loom_dns']['subdomain_whitelist']
+
+if subdomain_whitelist.nil? or subdomain_whitelist.include?(subdomain)
```
dereklwood: use || instead of 'or'... just a style thing

dereklwood:
also add a `default['loom_dns']['subdomain_whitelist'] = nil` to `attributes/default.rb` file

dereklwood:
style fix, and just add a default attribute (good practice), then :+1:

PR #585, opened 2014-09-22
Move apt.compile_time_update from template to cookbook

wolf31o2:
This allows it to be overridden, if necessary. Another cookbook can set `override['apt']['compile_time_update'] = false` to override.


wolf31o2:
This is an update to #583

albertshau:
:+1:

PR #586, opened 2014-09-22
Feature/no provider dns

wolf31o2:
This allows for not registering DNS when there is a provider-specific hostname. Currently, only AWS has externally-resolvable hostnames.


albertshau:
:+1:

PR #587, opened 2014-09-22
fix titles for ui pages so that the product name is used instead

albertshau:
of the company name, and the image types page is properly named.


wolf31o2:
:+1: LGTM

PR #590, opened 2014-09-22
Feature/aws vpc account

dereklwood:
- [x] making all the aws fields admin fields, so that defaults can be set by admin for vpc accounts.
- [x] updating default providers so that these fields show up in standalone


wolf31o2:
:+1: LGTM

PR #593, opened 2014-09-23
Feature/write creds

wolf31o2:
Write credentials to work directory when `@task['config']['ssh-auth']['identityfile']` and `@ssh_keyfile`


dereklwood:
:+1: lgtm

PR #597, opened 2014-09-24
Feature/ngui updates

wolf31o2:
Fix request URI and use full paths to node/server.js rather than letting npm take care of it...

### DIFF
--- a/ngui/server.js
+++ b/ngui/server.js
@@ -47,10 +47,12 @@ require('http-server')
           'Content-Type': 'application/json',
           'Cache-Control': 'no-store, must-revalidate'
         });
+        var cooprServerURI = process.env.COOPR_SERVER_URI || 'http://localhost:55044';
+        cooprServerURI += '/v2/';
         res.end(JSON.stringify({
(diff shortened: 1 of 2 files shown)


ngui/server.js
```diff
@@ -47,10 +47,12 @@ require('http-server')
           'Content-Type': 'application/json',
           'Cache-Control': 'no-store, must-revalidate'
         });
+        var cooprServerURI = process.env.COOPR_SERVER_URI || 'http://localhost:55044';
+        cooprServerURI += '/v2/';
```
nodesocket, line 51: Should `/v2` have a trailing slash?

wolf31o2, line 51: I tested with the trailing slash and it works... Checking javascript console and request URL doesn't have double `/` anywhere. Seems correct.

nodesocket, line 51: Cool, that means the requests are like `clusters` instead of `/clusters`. Generally I prefer to prefix requests with `/` so the root does not have a trailing slash. But that works. :+1:

PR #609, opened 2014-09-26
configtoformfields directive

gaarf:
import @nistala's directive, with tweaks, + test.

also moved existing directive tests to own directories.


gaarf:
@nistala @ajainarayanan please review.

coopr-ngui/app/js/directives/configtoformfields/configtoformfields-test.js
```diff
@@ -0,0 +1,102 @@
...
+    $compile = _$compile_;
+    scope = $rootScope.$new();
+  }));
+
+
+  it('without a config it renders nothing', function() {
+    var el = $compile('<my-configtoformfields />')(scope);
+    scope.$digest();
+
+    expect(el.find('label').length).toEqual(0);
```
ajainarayanan, line 17: Should we have something more like a state or a config initialized inside the directive that will more straight forward say that the config is not provided? I feel checking for label to be empty is not intuitive.

ajainarayanan:
Minor comments. But looks to good to be merged :+1:

ajainarayanan:
And @gaarf Could we have commit messages a bit more descriptive :) 

I generally follow this format - Adds/Removes/Updates X feature [Because of some brief reason].

That way we know what has been done so far in the PR by just looking at the commits. We have title and description to the PR but I feel this could add more to the readability. Up for suggestions.

gaarf:
granted, my commit messages are not very useful. It's about to get worse in next PR...

PR #613, opened 2014-09-30
Update imagetypes for Shellshock

wolf31o2:
Rackspace and Google have updated CentOS images... AWS, Joyent, and Rackspace have updated Ubuntu 12.04 images.


albertshau:
:+1:

PR #614, opened 2014-09-30
Remove (unconfigurable) disk_config option

wolf31o2:
This allows use of PVHVM images on Rackspace. We were setting the `disk_config` setting to `AUTO` when it was unconfigured. We had no way to configure this, yet, which made PVHVM images unusable. Removing this option until such time as we can add full support for the option.

More info:
http://www.rackspace.com/knowledge_center/article/choosing-a-virtualization-mode-pv-versus-pvhvm


albertshau:
:+1:

PR #615, opened 2014-10-01
Feature/ngui protractor sauce

gaarf:
ngui integration tests running in travis via https://saucelabs.com/ 

PR is against my branch https://github.com/caskdata/coopr/tree/gaarf - please review and merge #612 before this.

PR #616, opened 2014-10-01
Feature/ngui protractor sauce

gaarf:
same as #615 but against develop branch


ajainarayanan:
:+1: Good to be merged.

PR #617, opened 2014-10-01
disabling old ui tests since all new work should be going into the

albertshau:
new gui and because the tests are long and flaky.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,6 @@
 
     <modules>
         <module>coopr-server</module>
-        <module>coopr-integration-testing</module>
     </modules>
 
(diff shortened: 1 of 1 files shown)


wolf31o2:
If we're dropping the old UI tests from being run, can we just remove them, entirely?

wolf31o2:
Otherwise, LGTM on disabling...

albertshau:
was thinking we'll do that once the new ui tests are ready

albertshau:
we can also remove that top level pom at that point too...

wolf31o2:
Sounds good.

PR #619, opened 2014-10-01
COOPR-485: Update quickstart guide to mention the port to use when

albertshau:
uploading resources and the commands to use when using standalone
instead of installed packages.


wolf31o2:
:+1: LGTM

PR #624, opened 2014-10-02
COOPR-494 refactoring cluster callbacks to make sure writes cant

albertshau:
be done to the cluster store within the callback and to allow
looking up the profile of the account that owns the cluster. Also
restricting any reads of the cluster store to include only things
that are visible to the owner of the cluster.


poornachandra:
LGTM :+1:

PR #631, opened 2014-10-03
add builders for cluster template classes to make it clear what

albertshau:
fields are what and to make it easier to create classes since
many fields are optional. Also makes it easier to add fields
in the future.


poornachandra:
LGTM :+1:

PR #633, opened 2014-10-03
Add a persistent implementation of the credential store and

albertshau:
make the encryption/decryption happen outside of the class instead
of inside the memcache implementation so that it can be reused.


poornachandra:
LGTM :+1:

PR #635, opened 2014-10-06
Feature/rest client

amakogoncv:
Created interfaces for Coopr Rest Client API. @albertshau , please take a look. Interfaces contain all methods, are listed in the ticket https://issues.cask.co/browse/COOPR-493.  

**Note:** I added dependency to the _coopr-server_ project in to the _coopr-rest-client_ `pom.xml` for have ability to use some complicated request classes, like `ClusterCreateRequest`. So, I think, it would be good if we could move these request objects to the separate module, which will be accessible for all other parts of Coopr system. It could be _coopr-common_ module, or something like this. Please, let us to know, what do you think about this proposal.


coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,134 @@
...
+ * The client API for manage clusters.
+ */
+public interface ClusterClient {
+
+  /**
+   * Provides a summary of details about all clusters visible to a user.
+   *
+   * @return list of {@link co.cask.coopr.client.model.ClusterInfo} objects
+   */
+  List<ClusterInfo> getClusters();
```
albertshau: Don't think we should use separate objects than what the server uses. This should return a list of ClusterSummary

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  String createCluster(ClusterCreateRequest cluster);
+
+  /**
+   * Retrieves the status of a cluster by id.
+   *
+   * @param clusterId String value of a cluster id
+   * @return {@link co.cask.coopr.client.model.ClusterStatusInfo} object
+   */
+  ClusterStatusInfo getClusterStatus(String clusterId);
```
albertshau: this can return ClusterStatusResponse

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
```diff
@@ -0,0 +1,47 @@
...
+ * The client API for manage cluster templates.
+ */
+public interface ClusterTemplateClient {
+
+  /**
+   * Retrieves the list of all the cluster templates.
+   *
+   * @return List of {@link co.cask.coopr.client.model.ClusterTemplateInfo} objects
+   */
+  List<ClusterTemplateInfo> getAllClusterTemplates();
```
albertshau: similar to above, can just use ClusterTemplate class

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Retrieves details about a cluster template by the name.
+   *
+   * @return {@link co.cask.coopr.client.model.ClusterTemplateInfo} object
+   */
+  ClusterTemplateInfo getClusterTemplate(String name);
```
albertshau: ClusterTemplate instead of this class

coopr-rest-client/src/main/java/co/cask/coopr/client/HardwareTypeClient.java
```diff
@@ -0,0 +1,47 @@
...
+ * The client API for manage the hardware specifications.
+ */
+public interface HardwareTypeClient {
+
+  /**
+   * Retrieves the list of all the hardware types.
+   *
+   * @return List of {@link co.cask.coopr.client.model.HardwareTypeInfo} objects
+   */
+  List<HardwareTypeInfo> getAllHardwareTypes();
```
albertshau: HardwareTypeInfo -> HardwareType

coopr-rest-client/src/main/java/co/cask/coopr/client/model/AbstractJsonInfo.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+package co.cask.coopr.client.model;
+
+import com.google.gson.JsonObject;
+
+/**
+ * Abstract info class with json content.
+ */
+public class AbstractJsonInfo implements JsonInfo {
```
albertshau: Don't think there is a need for this class and its children, we can just use the classes that are already defined in the server project (with an aim to separate these out later).

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
albertshau: Javadoc should describe what is returned if no such template exists. Assume the plan is to return null?

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
albertshau: Javadoc should describe what is returned if no templates exist. To be consistent with the rest api, I think it should be an empty list.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
```diff
@@ -0,0 +1,47 @@
...
+package co.cask.coopr.client;
+
+import co.cask.coopr.client.model.ClusterTemplateInfo;
+
+import java.util.List;
+
+/**
+ * The client API for manage cluster templates.
+ */
+public interface ClusterTemplateClient {
```
albertshau: I think we can combine the template, provider, service, hardware, and image clients into a single client since anybody who can get one type of object can get any of the other types of objects.

coopr-rest-client/src/main/java/co/cask/coopr/client/PluginClient.java
```diff
@@ -0,0 +1,153 @@
...
+   * Retrieves a mapping of all resources of the given type for the given provider type.
+   * Method can optionally contain a 'status' parameter, whose value is one of 'active', 'inactive', 'staged', or
+   * 'recalled' to filter the results to only contain resource that have the given status.
+   *
+   * @param id Id of the provider type that has the resources
+   * @param resourceType Type of resource to get
+   * @param status Status of the resources to get. If null, resources of any status are returned
+   * @return Immutable map of resource name to resource metadata
+   */
+  Map<String, Set<ResourceMetaInfo>> getProviderTypeResources(String id, String resourceType, String status);
```
albertshau: status can use the ResourceStatus enum

coopr-rest-client/src/main/java/co/cask/coopr/client/PluginClient.java
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Stage a particular resource version, which means that version of the resource will get pushed to provisioners
+   * on the next sync call. Staging one version recalls other versions of the resource.
+   *
+   * @param id Id of the automator type that has the resources
+   * @param resourceType Type of resource to stage
+   * @param resourceName Name of the resource to stage
+   * @param version Version of the resource to stage
+   */
+  void stageAutomatorTypeModule(String id, String resourceType, String resourceName, String version);
```
albertshau: "module" was an older term that got changed to "resource", so we should use "resource" everywhere instead of "module".  (Seems we missed some places in the server code, will need to rename those too).

albertshau:
Yep, we should clean up the server project and break out a lot of these objects into another module.  However, that is something that can be done in a separate pull request so for the time being we should just depend on the server and use the classes there.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  List<ClusterInfo> getClusters();
+
+  /**
+   * Provides full details about a cluster by id.
+   *
+   * @param clusterId String value of a cluster id
+   * @return {@link co.cask.coopr.client.model.ClusterInfo} object
+   */
+  ClusterInfo getCluster(String clusterId);
```
albertshau: this can return ClusterDetails (once #638 is merged)

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterTemplateClient.java
amakogoncv: Maybe will be better to throw some Runtime Exception the same, as in the CDAP Rest Stream Client?

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
amakogoncv: Ok, for now, I replaced `ClusterInfo` with `JsonObject` class.

coopr-rest-client/src/main/java/co/cask/coopr/client/PluginClient.java
amakogoncv: Ok, it is clear. Fixed it.

coopr-rest-client/src/main/java/co/cask/coopr/client/model/AbstractJsonInfo.java
amakogoncv: Removed this class and its children and replaced them with appropriate specification classes from the coopr-server. 
In two methods replaced them with `JsonObject`:
`JsonObject getCluster(String clusterId);`
`JsonObject getClusterConfig(String clusterId);`
Because didn't find some specification classes, which could be used.

albertshau:
travis tests are failing due to Sauce settings.  Once #639 is merged, you can merge develop and the tests should pass

coopr-rest-client/src/main/java/co/cask/coopr/client/AdminClient.java
```diff
@@ -0,0 +1,128 @@
...
+import co.cask.coopr.spec.HardwareType;
+import co.cask.coopr.spec.ImageType;
+import co.cask.coopr.spec.Provider;
+import co.cask.coopr.spec.service.Service;
+import co.cask.coopr.spec.template.ClusterTemplate;
+
+import java.util.List;
+
+/**
+ * The client API for manage admin defined entities..
```
albertshau: extra period

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Sync the cluster template of the cluster to the current version of the cluster template by the cluster id.
+   *
+   * @param clusterId String value of a cluster id
+   */
+  void syncClusterTemplate(String clusterId);
+
+  /**
+   * Changes a cluster expiration timestamp.
```
albertshau: I think 'set' would be a term for this than 'change', because 'change' could be interpreted as incrementing the expire time.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @param clusterId String value of a cluster id
+   * @param expireTime long value of expiration time in milliseconds
```
albertshau: timestamp in milliseconds to set the expire time to

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   */
+  void changeClusterExpireTime(String clusterId, long expireTime);
```
albertshau: change -> set

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   */
+  void deleteCluster(String clusterId);
+
+  /**
+   * Creates new Cluster according to the specified parameters.
+   *
+   * @param cluster {@link co.cask.coopr.http.request.ClusterCreateRequest} object
+   * @return new Cluster id
+   */
+  String createCluster(ClusterCreateRequest cluster);
```
albertshau: createRequest or clusterCreateRequest would be a better name for this variable than cluster

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   * Provides a summary of details about all clusters visible to a user.
+   * If there are no visible clusters, returns empty list.
+   *
+   * @return list of {@link co.cask.coopr.cluster.ClusterSummary} objects
+   */
+  List<ClusterSummary> getClusters();
+
+  /**
+   * Provides full details about a cluster by id.
+   *
```
albertshau, line 48: maybe add a TODO: comment here to replace JsonObject with ClusterDetails once its merged so we don't forget.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+public interface ClusterClient {
+
+  /**
+   * Provides a summary of details about all clusters visible to a user.
+   * If there are no visible clusters, returns empty list.
+   *
+   * @return list of {@link co.cask.coopr.cluster.ClusterSummary} objects
+   */
+  List<ClusterSummary> getClusters();
```
albertshau: general comment about these calls is what happens under error conditions, for example if the API is not accessible?  Seems like we should throw an exception in that scenario, which would require a change to all these methods.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   */
+  ClusterStatusResponse getClusterStatus(String clusterId);
+
+  /**
+   * Retrieves the configuration of a cluster by id.
+   *
+   * @param clusterId String value of a cluster id
+   * @return {@link com.google.gson.JsonObject} object
+   */
+  JsonObject getClusterConfig(String clusterId);
```
albertshau: client should also have a method for setting the cluster config, with clusterId and ClusterConfigureRequest as input

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   * @return {@link com.google.gson.JsonObject} object
+   */
+  JsonObject getCluster(String clusterId);
+
+  /**
+   * Deletes specified cluster by id.
+   *
+   * @param clusterId String value of a cluster id
+   */
+  void deleteCluster(String clusterId);
```
albertshau: We should have another method for delete that takes a ClusterOperationRequest as an additional arg.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+   */
+  void changeClusterExpireTime(String clusterId, long expireTime);
+
+  /**
+   * Starts a specific service on a cluster.
+   *
+   * @param clusterId String value of a cluster id
+   * @param serviceId String value of a specific service id
+   */
+  void startServiceOnCluster(String clusterId, String serviceId);
```
albertshau: We should have another method for this that takes a ClusterOperationRequest as an additional arg.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Stops a specific service on a cluster.
+   *
+   * @param clusterId String value of a cluster id
+   * @param serviceId String value of a specific service id
+   */
+  void stopServiceOnCluster(String clusterId, String serviceId);
```
albertshau: We should have another method for this that takes a ClusterOperationRequest as an additional arg.

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * Restarts a specific service on a cluster.
+   *
+   * @param clusterId String value of a cluster id
+   * @param serviceId String value of a specific service id
+   */
+  void restartServiceOnCluster(String clusterId, String serviceId);
```
albertshau: We should have another method for this that takes a ClusterOperationRequest as an additional arg.

albertshau:
a few comments then we should be good to merge

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
amakogoncv: I think, IOException could be used for this. What do you think about this?

albertshau: sounds good to me

coopr-rest-client/src/main/java/co/cask/coopr/client/ClusterClient.java
amakogoncv: Should I modified existing method or create another one?

amakogoncv:
Made changes according to the Albert's comments. @albertshau please, take a look and merge this if you have no other comments.

PR #636, opened 2014-10-06
adding builders for service related classes to make it easier

albertshau:
to add fields in the future without changing constructors all
over the test code.


poornachandra:
LGTM :+1:

PR #637, opened 2014-10-06
Simply log and continue on missing servers

wolf31o2:
Catch another way Joyent could error on a missing server.


albertshau:
:+1:

PR #638, opened 2014-10-07
COOPR-496 add links to the cluster template that can point to

albertshau:
various useful service UIs. Services and cluster templates have
a links field that consists of an array of links, where a link
is a label and a url. The url can contain macros. When a cluster
is fetched from the rest api, links in the cluster template are
expanded and placed at the top level of the returned object.
In addition, for every node in the cluster, the links for all
services on the node are taken, expanded, and combined into a list
of links for the node.

### DIFF
--- a/coopr-server/src/main/java/co/cask/coopr/cluster/Node.java
+++ b/coopr-server/src/main/java/co/cask/coopr/cluster/Node.java
@@ -15,6 +15,8 @@
  */
 package co.cask.coopr.cluster;
 
+import co.cask.coopr.macro.Expander;
+import co.cask.coopr.spec.Link;
 import co.cask.coopr.spec.service.Service;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
@@ -23,6 +25,8 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
@@ -31,14 +35,14 @@ import java.util.Set;
 /**
  * Represents a machine in a cluster.
  */
-public final class Node implements Comparable<Node> {
-  private static final String IPADDRESS_KEY = "ipaddress";
+public class Node implements Comparable<Node> {
   private final String id;
   private final String clusterId;
   private final Set<Service> services;
   private final NodeProperties properties;
   private final List<Action> actions;
   private final JsonObject provisionerResults;
+  private List<Link> links;
 
   /**
    * Node status.
@@ -157,6 +161,27 @@ public final class Node implements Comparable<Node> {
     for (Map.Entry<String, JsonElement> entry : results.entrySet()) {
       this.provisionerResults.add(entry.getKey(), entry.getValue());
     }
+  }
+
+  /**
+   * Get all service links on the node, combine them all, and expand any macros in them. Only useful for display
+   * purposes.
+   */
+  public void populateLinks(Cluster cluster, Set<Node> nodes) {
+    // take links from the services on this node, expand any self macros that may be there, and combine them all
+    ImmutableList.Builder<Link> linksBuilder = ImmutableList.builder();
+    for (Service service : services) {
+      for (Link link : service.getLinks()) {
+        try {
+          // The service link may have macros like %host.self% that should get expanded.
+          linksBuilder.add(new Link(link.getLabel(), Expander.expand(link.getUrl(), cluster, nodes, this)));
+        } catch (Exception e) {
+          // if we couldn't expand the macro, just use the original string
+          linksBuilder.add(link);
+        }
+      }
+    }
+    this.links = linksBuilder.build();
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


wolf31o2:
Could you provide an example of what "links" may look like for the `hadoop-hdfs-namenode` service?

Also, if no services had any links, but there were links in a clustertemplate, only those links would be returned at the top-level and none would be returned per-node, correct?

albertshau:
yup so you might have for a service:

```
{
    "name":"hadoop-hdfs-namenode",
    "links":[
        {
            "label": "HDFS Namenode",
            "url": "http://%ip.access_v4.self%:50070"
        }
    ],
    ...
}
```

and then for a template you might have:

```
{
    "name":"hadoop-distributed",
    "links":[
        {
            "label": "HDFS Namenode",
            "url": "http://%ip.access_v4.service.hadoop-hdfs-namenode%:50070"
        },
        {
            "label": "HDFS Resource Manager",
            "url": "http://%ip.access_v4.service.hadoop-yarn-resourcemanager%:8088"
        },
        {
            "label": "HBase Master",
            "url": "http://%ip.access_v4.service.hbase-master%:60010"
        },
    ],
    ...
}
```

albertshau:
and then these would get expanded on a `GET /clusters/<id>` to:

```
{
    "name":"my-cluster-name",
    "links": [
        {
            "label": "HDFS Namenode",
            "url": "http://1.2.3.4:50070"
        },
        {
            "label": "HDFS Resource Manager",
            "url": "http://1.2.3.4:8088"
        },
        {
            "label": "HBase Master",
            "url": "http://1.2.3.4:60010"
        }
    ],
    ...
    "nodes": [
        {
            // assume this is the namenode
            "id": ...,
            "clusterId": ...,
            "links": [
                 {
                    "label": "HDFS Namenode",
                    "url": "http://1.2.3.4:50070"
                },
                {
                    "label": "HDFS Resource Manager",
                    "url": "http://1.2.3.4:8088"
                },
                {
                    "label": "HBase Master",
                    "url": "http://1.2.3.4:60010"
                }
            ],
            ...
        },
        ...    
    ]
}
```

wolf31o2:
Cool. Looks as I expected. Thanks.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/ServiceCodec.java
```diff
@@ -65,8 +68,9 @@ public Service deserialize(JsonElement json, Type type, JsonDeserializationConte
     Map<ProvisionerAction, ServiceAction> actions = Collections.emptyMap();
     if (provisioner != null) {
       actions = context.deserialize(provisioner.get("actions"),
-                                   new TypeToken<Map<ProvisionerAction, ServiceAction>>(){}.getType());
+                                    new TypeToken<Map<ProvisionerAction, ServiceAction>>() {}.getType());
```
poornachandra: new TypeToken<Map<ProvisionerAction, ServiceAction>>() {}.getType() can be a constant.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/ServiceCodec.java
```diff
@@ -65,8 +68,9 @@ public Service deserialize(JsonElement json, Type type, JsonDeserializationConte
...
     }
+    Set<Link> links = context.deserialize(jsonObj.get("links"), new TypeToken<Set<Link>>() {}.getType());
```
poornachandra: new TypeToken<Set<Link>>() {}.getType() can be a constant.

coopr-server/src/main/java/co/cask/coopr/cluster/Node.java
```diff
@@ -31,14 +35,14 @@
...
-public final class Node implements Comparable<Node> {
-  private static final String IPADDRESS_KEY = "ipaddress";
+public class Node implements Comparable<Node> {
   private final String id;
   private final String clusterId;
   private final Set<Service> services;
   private final NodeProperties properties;
   private final List<Action> actions;
   private final JsonObject provisionerResults;
+  private List<Link> links;
```
poornachandra, line 45: links is not exposed by any public method. What will links be used for?

coopr-server/src/main/java/co/cask/coopr/codec/json/current/ClusterTemplateCodec.java
```diff
@@ -62,6 +66,7 @@ public ClusterTemplate deserialize(JsonElement json, Type type, JsonDeserializat
       .setCompatibilities(context.<Compatibilities>deserialize(jsonObj.get("compatibility"), Compatibilities.class))
       .setConstraints(context.<Constraints>deserialize(jsonObj.get("constraints"), Constraints.class))
       .setAdministration(context.<Administration>deserialize(jsonObj.get("administration"), Administration.class))
+      .setLinks(context.<Set<Link>>deserialize(jsonObj.get("links"), new TypeToken<Set<Link>>() {}.getType()))
```
poornachandra: new TypeToken<Set<Link>>() {}.getType() can be a constant.

poornachandra:

coopr-server/src/main/java/co/cask/coopr/cluster/Node.java
albertshau, line 45: will add a getter.  Not currently used, but may be used by some cluster callback for example.

PR #641, opened 2014-10-07
Write out SSH file if ssh_keyfile

wolf31o2:
Writes out task-specific SSH files when `@task['config']['provider']['provisioner']['ssh_keyfile']` is set and clean up after ourselves.


albertshau:
:+1:

PR #643, opened 2014-10-09
COOPR-513 adding label as a field to providers, hardwaretypes,

albertshau:
imagetypes, services, and cluster templates. Also doing some
refactoring so they all share the same code for common fields.


coopr-server/src/main/java/co/cask/coopr/codec/json/current/ImageTypeCodec.java
```diff
@@ -30,31 +28,18 @@
 /**
  * Codec for serializing/deserializing a {@link ImageType}.
  */
-public class ImageTypeCodec extends AbstractCodec<ImageType> {
+public class ImageTypeCodec extends BaseAdminEntityCodec<ImageType> {
+  private static final Type PROVIDERMAP_TYPE = new TypeToken<Map<String, Map<String, String>>>() {}.getType();
```
sreevatsanraman, line 32: Open a jira to fix the providermap_type - would be better to have a a Map<String, ProviderInfo> where the Type ProviderInfo encodes the provider information instead of a Map of a Map.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/HardwareTypeCodec.java
```diff
@@ -30,31 +28,18 @@
 /**
  * Codec for serializing/deserializing a {@link HardwareType}.
  */
-public class HardwareTypeCodec extends AbstractCodec<HardwareType> {
+public class HardwareTypeCodec extends BaseAdminEntityCodec<HardwareType> {
+  private static final Type PROVIDERMAP_TYPE = new TypeToken<Map<String, Map<String, String>>>() {}.getType();
```
sreevatsanraman, line 32: Same comment as mentioned earlier.

sreevatsanraman:
LGTM :+1:  please open a jira to clean up the provider info.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/ImageTypeCodec.java
albertshau, line 32: opened  COOPR-515

PR #644, opened 2014-10-09
COOPR-324 make sure all system generated hostnames have labels

albertshau:
that do not exceed 63 characters.
Also lowercasing the entire hostname for consistency since they
should be case insensitive anyway.


dereklwood:
:+1: lgtm

PR #646, opened 2014-10-09
only check for disk vdb for now since thats all we operate on

dereklwood:
short-term fix for joyent mounting issue. only check for /vdb for now, since thats the only disk we attempt to remount. longer term we should individually operate on vdb and then vdb1


wolf31o2:
:+1: LGTM

PR #647, opened 2014-10-09
include common entity fields in the get all clusters call.

albertshau:


wolf31o2:
:+1: LGTM (but it's late and Java, so....)

PR #648, opened 2014-10-09
Feature/rest client

amakogoncv:
Start to implement Client Interfaces, submitted before according to the https://issues.cask.co/browse/COOPR-501 jira ticket.
@albertshau , take a look, please, is such structure ok for you? If you have some recommendations, please, leave these here. For now, only `AdminRestClient` and `ClusterRestClient` are implemented. I'm also going to finish with another clients and created unit tests in next two days.


coopr-rest-client/src/main/java/co/cask/coopr/client/ClientManager.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package co.cask.coopr.client;
+
+import java.io.Closeable;
+
+/**
+ * The API for manage all existing clients with a single instance.
```
albertshau: The API for managing singletons of all existing clients.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+    super(config, httpClient);
+  }
+
+  @Override
+  public List<ClusterSummary> getClusters() throws IOException {
+    return getAll(CLUSTERS_URL_SUFFIX);
+  }
+
+  @Override
+  public JsonObject getCluster(String clusterId) throws IOException {
```
albertshau: If you merge develop, this method and its parent methods can now be changed to return the ClusterDetails object

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+    try {
+      RestClient.responseCodeAnalysis(httpResponse);
+    } finally {
+      httpResponse.close();
+    }
+  }
+
+  @Override
+  public Set<String> getClusterServices(String clusterId) throws IOException {
+    HttpGet getRequest = new HttpGet(getBaseURL().resolve(
```
albertshau: how come this can't use the RestClient's getAll method?

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+      newClusterId = jsonContent.get(CLUSTER_ID_ATTRIBUTE_NAME).getAsString();
+    } finally {
+      httpResponse.close();
+    }
+    return newClusterId;
+  }
+
+  @Override
+  public ClusterStatusResponse getClusterStatus(String clusterId) throws IOException {
+    HttpGet getRequest = new HttpGet(getBaseURL().resolve(
```
albertshau: can this use the superclass' getSingle method?

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+  public void deleteCluster(String clusterId) throws IOException {
+    deleteCluster(clusterId, null);
+  }
+
+  @Override
+  public void deleteCluster(String clusterId, ClusterOperationRequest clusterOperationRequest) throws IOException {
+    HttpDeleteWithContent deleteRequest =
+      new HttpDeleteWithContent(getBaseURL().resolve(String.format("/%s/clusters/%s", getVersion(), clusterId)));
+    if (clusterOperationRequest != null) {
+      StringEntity stringEntity = new StringEntity(GSON.toJson(clusterOperationRequest), Charsets.UTF_8);
```
albertshau: You could add code to the base class for setting a request body, which could then be re-used in multiple places instead of explicitly handled in each of the calls that require it.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+      RestClient.responseCodeAnalysis(httpResponse);
+    } finally {
+      httpResponse.close();
+    }
+  }
+
+  @Override
+  public String createCluster(ClusterCreateRequest clusterCreateRequest) throws IOException {
+    HttpPost postRequest = new HttpPost(getBaseURL().resolve(String.format("/%s/clusters", getVersion())));
+    if (clusterCreateRequest != null) {
```
albertshau: unlike the other cluster operations, creating a cluster always requires a request body, so this should never be null.  We can use a Precondition to check this.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
albertshau: o i see it doesn't fit the same pattern.  Maybe add a method to the parent where you supply the path so that the rest of the code can be re-used here and a few other of the get calls

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+   * @return the base URL of Rest Service API
+   */
+  public URI getBaseURL() {
+    return baseUrl;
+  }
+
+  /**
+   * @return the version of Rest Service API
+   */
+  public String getVersion() {
```
albertshau: this can be hidden inside the class and put in the baseUrl since every API call starts with it.  That way clients don't have to know about version and don't have to give it in their paths.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+    } finally {
+      httpResponse.close();
+    }
+    return clusterServices != null ? clusterServices : new HashSet<String>();
+  }
+
+  @Override
+  public void syncClusterTemplate(String clusterId) throws IOException {
+    HttpPost postRequest =
+      new HttpPost(getBaseURL().resolve(String.format("/%s/clusters/%s/clustertemplate/sync", getVersion(),
```
albertshau: in general, I think it's better to avoid making the clients include the version in their path.  It can be put into the base path in the parent.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+      httpResponse.close();
+    }
+    return clusterDetails;
+  }
+
+  @Override
+  public void setClusterConfig(String clusterId, ClusterConfigureRequest clusterConfigureRequest) throws IOException {
+    HttpPut putRequest = new HttpPut(getBaseURL().resolve(
+      String.format("/%s/clusters/%s/config", getVersion(), clusterId)));
+    if (clusterConfigureRequest != null) {
```
albertshau: a request to set the cluster config must also contain a request body, so this should never be null

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+                                            new TypeToken<ClusterStatusResponse>() { }.getType());
+    } finally {
+      httpResponse.close();
+    }
+    return clusterStatusResponse;
+  }
+
+  @Override
+  public JsonObject getClusterConfig(String clusterId) throws IOException {
+    HttpGet getRequest = new HttpGet(getBaseURL().resolve(
```
albertshau: can we use the parent's getSingle method here?

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
```diff
@@ -0,0 +1,294 @@
...
+    } finally {
+      httpResponse.close();
+    }
+  }
+
+  @Override
+  public void setClusterExpireTime(String clusterId, long expireTime) throws IOException {
+    HttpPost postRequest = new HttpPost(getBaseURL().resolve(String.format("/%s/clusters/%s", getVersion(),
+                                                                           clusterId)));
+    StringEntity entity = new StringEntity(GSON.toJson(ImmutableMap.of(EXPIRE_TIME_ATTRIBUTE_NAME, expireTime)));
```
albertshau, line 153: this type of request body setting code shows up in multiple places, should be able to refactor to use the same code.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+  public RestClient(RestClientConnectionConfig config, CloseableHttpClient httpClient) {
+    this.config = config;
+    this.baseUrl = URI.create(String.format("%s://%s:%d", config.isSSL() ? HTTPS_PROTOCOL : HTTP_PROTOCOL,
+                                            config.getHost(), config.getPort()));
+    this.version = config.getVersion();
+    this.httpClient = httpClient;
+  }
+
+  /**
+   * Method for execute HttpRequest with authorized headers, if need.
```
albertshau: "for executing", "if needed"

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+   *
+   * @param request {@link org.apache.http.client.methods.HttpRequestBase} initiated http request with entity, headers,
+   *                                                                      request uri and all another required
```
albertshau: "all other"

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+   *                                                                      properties for successfully request
```
albertshau: can leave out "for successfully request"

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+   * @return {@link org.apache.http.client.methods.CloseableHttpResponse} as a result of http request execution
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  public CloseableHttpResponse execute(HttpRequestBase request) throws IOException {
+    if (StringUtils.isNotEmpty(config.getUserId())) {
```
albertshau: the user header is always required (at least today).  Checking the user should be done once in the constructor instead of for each execute call

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+      request.setHeader(COOPR_USER_ID_HEADER_NAME, config.getUserId());
+    }
+    if (StringUtils.isNotEmpty(config.getTenantId())) {
```
albertshau: similar to above, the tenant header is always required and should be checked once in the constructor instead of for each execute call

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+      request.setHeader(COOPR_TENANT_ID_HEADER_NAME, config.getTenantId());
+    }
+    //TODO: Remove API Key if it's not used anymore
+    if (StringUtils.isNotEmpty(config.getAPIKey())) {
```
albertshau: should be checked just once instead of with each call.  This header is currently not checked, though it will be in the future.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+    LOG.debug("Execute Http Request: {}", request);
+    return httpClient.execute(request);
+  }
+
+  /**
+   * Utility method for analysis http response status code and throw appropriate Java API Exception.
+   *
+   * @param response {@link org.apache.http.HttpResponse} http response
+   */
+  public static void responseCodeAnalysis(HttpResponse response) {
```
albertshau: it's generally better for method names to be verbs rather than nouns, so "analyzeResponseCode" would be better.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,177 @@
...
+    //TODO: Remove API Key if it's not used anymore
+    if (StringUtils.isNotEmpty(config.getAPIKey())) {
+      request.setHeader(COOPR_API_KEY_HEADER_NAME, config.getAPIKey());
+    }
+    LOG.debug("Execute Http Request: {}", request);
+    return httpClient.execute(request);
+  }
+
+  /**
+   * Utility method for analysis http response status code and throw appropriate Java API Exception.
```
albertshau: "analyzing", "and throwing the appropriate"

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClientManager.java
```diff
@@ -0,0 +1,192 @@
...
+        LOG.error("Failed to init SSL context: {}", e);
+      } catch (NoSuchAlgorithmException e) {
+        LOG.error("Failed to get instance of SSL context: {}", e);
+      }
+    }
+    this.httpClient = HttpClients.custom().setConnectionManager(createConnectionManager()).build();
+  }
+
+  /**
+   * Create builder for build RestClientManager instance.
```
albertshau: "for build" -> "for building a"

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClientManager.java
```diff
@@ -0,0 +1,192 @@
...
+   * @param port coopr server port
+   * @return {@link Builder} Builder instance
+   */
+  public static Builder builder(String host, int port) {
+    return new Builder(host, port);
+  }
+
+  @Override
+  public AdminClient getAdminClient() {
+    if (adminClient == null) {
```
albertshau: is there any cost to creating a client? If not it's better to make these clients final and set them in the constructor instead of lazy creating them.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+
+/**
+ * Utility methods for working with HttpClient.
+ */
+public final class RestUtil {
+
+  private RestUtil() { }
```
albertshau, line 38: seems like we can remove this constructor

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+  public static Registry<ConnectionSocketFactory> getRegistryWithDisabledCertCheck()
```
albertshau, line 49: javadoc missing

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/exception/HttpFailureException.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package co.cask.coopr.client.rest.exception;
+
+/**
+ * Runtime exception thrown when execution of HTTP request is considered a failure.
+ */
+public class HttpFailureException extends RuntimeException {
```
albertshau, line 22: It doesn't seem like this should be a runtime exception.  They seem like they should be things you want to force people to catch and handle in their code.

coopr-rest-client/pom.xml
```diff
@@ -12,11 +12,20 @@
...
+
   <dependencies>
     <dependency>
       <groupId>co.cask</groupId>
       <artifactId>coopr</artifactId>
-      <version>0.9.9-SNAPSHOT</version>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-lang</groupId>
```
albertshau: seems like the only thing that needed this was the not implemented yet exception.  If that's the case can we throw something else like an UnsupportedOperationException instead of adding another dependency?

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestUtil.java
amakogoncv, line 38: This constructor is needed to avoid a creating of instances from this class, because this class includes only static method(s).
You could take a look http://www.informit.com/articles/article.aspx?p=1216151&seqNum=4

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/exception/HttpFailureException.java
amakogoncv, line 22: I created and used this exception by analogy with rest clients, we implemented in the cdap project. Also there are some similar exceptions in the `javax.ws.rs` package (like `BadRequestException`, `NotFoundException` etc) and these exceptions are also runtime. 
But if you want, I could change it to the checked exception.

coopr-rest-client/pom.xml
amakogoncv: I also used `StringUtils` from this lib, but I investigated that I can replace it by the `Strings` from the `guava` lib. But, anyway, in the parent `pom.xml` file there is apache commons dependency:

```
        <dependency>
            <groupId>org.apache.commons</groupId>
            <artifactId>commons-lang3</artifactId>
            <version>3.3</version>
        </dependency>
```

so, I left `UnsupportedOperationException` exception.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/ClusterRestClient.java
amakogoncv: By the same reason, as you have noticed below. But, I added the same `getAll` method with the full URL as parameter and changed returned type for this method to the `List` instead of `Set` and then had ability to reuse `getAll` method.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/exception/HttpFailureException.java
albertshau, line 22: hm ok let's leave it as it is then

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestUtil.java
albertshau, line 38: ah got it

coopr-rest-client/src/test/java/co/cask/coopr/client/Entities.java
```diff
@@ -0,0 +1,633 @@
...
+import com.google.gson.JsonObject;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A bunch of example json strings for testing purposes.
+ */
+public class Entities {
```
albertshau: is this a copy? you should be able to use the one in the server.

coopr-rest-client/pom.xml
```diff
@@ -12,11 +12,22 @@
...
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>co.cask</groupId>
       <artifactId>coopr</artifactId>
-      <version>0.9.9-SNAPSHOT</version>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
```
albertshau, line 51: isn't this already in the parent?

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,197 @@
...
+      httpResponse.close();
+    }
+    return resultList != null ? resultList : new ArrayList<T>();
+  }
+
+  protected <T> T getSingle(String urlSuffix, String name, TypeToken<T> typeToken) throws IOException {
+    return getSingle(URI.create(String.format("%s/%s/%s", baseUrl, urlSuffix, name)), typeToken);
+  }
+
+  protected <T> T getSingle(URI url, TypeToken<T> typeToken) throws IOException {
```
albertshau: it will be cleaner to have this take a Type instead of TypeToken. Then the caller can just send in ClusterDetails.class instead of a type token.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
```diff
@@ -0,0 +1,197 @@
...
+      default:
+        throw new UnsupportedOperationException("Operation is not supported by the Coopr Server");
+    }
+  }
+
+  protected <T> List<T> getAll(String urlSuffix, TypeToken<List<T>> typeToken) throws IOException {
+    return getAll(URI.create(String.format("%s/%s", baseUrl, urlSuffix)), typeToken);
+  }
+
+  protected <T> List<T> getAll(URI url, TypeToken<List<T>> typeToken) throws IOException {
```
albertshau: I believe you should be able to take a Type instead of TypeToken and make the list type token inside the method. That would make it simpler for callers, as they can just pass in Service.class or whatever they want.

coopr-rest-client/src/test/java/co/cask/coopr/client/rest/handler/ClusterTemplatesHandler.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Cluster Templates test handler.
+ */
+public class ClusterTemplatesHandler extends AbstractAdminHandler<ClusterTemplate> {
+
+  private static final String GET_ALL_URL = "/v2/clustertemplates";
+
+  @Override
+  public List<ClusterTemplate> getAll() {
+    return Arrays.asList(Entities.ClusterTemplateExample.HADOOP_DISTRIBUTED, Entities.ClusterTemplateExample.HDFS,
```
albertshau: could also use Lists.newArrayList(...)

coopr-rest-client/src/test/java/co/cask/coopr/client/rest/AdminRestClientTest.java
```diff
@@ -0,0 +1,75 @@
...
+    adminClient = clientManager.getAdminClient();
+    adminClient.getAllClusterTemplates();
+  }
+
+  @Test
+  public void testGetClusterTemplateSuccess() throws IOException {
+    ClusterTemplate result = adminClient.getClusterTemplate(HADOOP_DISTRIBUTED_NAME);
+    assertTrue(result.equals(Entities.ClusterTemplateExample.HADOOP_DISTRIBUTED));
+  }
+
```
albertshau, line 83: would be better to cover more than just these 3 methods.  At least the calls that don't fit into the getall, getsingle, delete pattern. For example, the call to restart a service or to set the expire time of a cluster.

albertshau:
a few comments, then good to merge

coopr-rest-client/pom.xml
amakogoncv, line 51: If you mean  `<groupId>org.apache.httpcomponents</groupId>`, yes such dependency already exists in the parent `pom.xml`. But need also separate one with the testing scope.

coopr-rest-client/src/test/java/co/cask/coopr/client/Entities.java
amakogoncv: Yes. Done. Found the way to reuse it.

coopr-rest-client/src/test/java/co/cask/coopr/client/rest/AdminRestClientTest.java
amakogoncv, line 83: Yes, sure. I didn't have enough of time for this yesterday.

coopr-rest-client/src/main/java/co/cask/coopr/client/rest/RestClient.java
amakogoncv: Got it.

amakogoncv:
@albertshau , I have updated code according to your comments above. Unfortunately, I had time to write only positive test cases for `ClusterRestClient` today. I'm going to add negative cases also, but would be good if you merge this PR today, if you has not any comments. This is need to make this API available to the CLI Client, which is implemented in the another brunch at the moment.

albertshau:
lgtm. one small comment but will merge this first and leave it for another PR

PR #649, opened 2014-10-09
COOPR-516 Move a cluster to the 'terminated' state instead of

albertshau:
'incomplete' if all create tasks failed in this manner.


coopr-server/src/main/java/co/cask/coopr/scheduler/JobScheduler.java
```diff
@@ -178,8 +184,15 @@ public void run() {
             }
             clusterStore.writeClusterJob(job);
           } else if (inProgressTasks == 0) {
-            // Job failed and no in progress tasks remaining, update cluster status
-            taskService.failJobAndSetClusterStatus(job, cluster);
+            // special case: if all tasks were create tasks and all of them failed before they created anything,
+            // set the cluster state to 'terminated' instead of letting it go to 'incomplete'.
+            if (createFailedTasks == currentStage.size()) {
```
poornachandra: It would be safer to check if none of the creates succeeded here to make sure there are no edge cases.

poornachandra:
Done with review, please address the comment.

poornachandra:
LGTM :+1:

PR #650, opened 2014-10-10
COOPR-497 - Split config setting 

artemmakogon:


artemmakogon:
Branch is ready for review.

coopr-server/src/test/java/co/cask/coopr/common/conf/ConfigurationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.collect.MapDifference;
+import com.google.common.collect.Maps;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * ConfigurationTest checks that the obtaining properties from the coopr-security.xml is working correctly.
```
awholegunch: Checks that obtaining properties from coopr-security.xml works correctly.

coopr-server/distribution/etc/coopr/conf.dist/coopr-security.xml.example
```diff
@@ -0,0 +1,31 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+-->
+<configuration>
+  <property>
+    <name>server.db.user</name>
+    <value>coopr</value>
+    <description>mysql user</description>
```
wolf31o2: I know that you only moved this from the original example file, but this shouldn't mention MySQL. It should be "Database user" and "Database password" since Coopr can use any JDBC-compatible database.

artemmakogon:
Fixed. Thank you.

albertshau:
:+1:

PR #651, opened 2014-10-10
Use status 201 for Unauthorized

wolf31o2:
See caskdata/coopr#649 and COOPR-516 for background information. The idea is to allow the server to know why the task failed, so it can take an appropriate action based off the context of the cluster job.


dereklwood:
:+1:

PR #652, opened 2014-10-11
Adds popup mode to 'my-thing-picker'

ajainarayanan:
Example usage 

``` html
<my-thing-picker
            data-model="model.dependencies.provides"
            data-allow-rm="true"
            data-thing-name="Service"
            data-list-inline="true"
            data-freetext-mode="true"
          >
          </my-thing-picker>
```

PR #653, opened 2014-10-13
Feature/ngui services

gaarf:
this + #652 = ngui feature parity with previous ui


ajainarayanan:
:+1: LGTM. We just need some offline discussion about the focus manager stuff.

gaarf:
Thanks @ajainarayanan !

PR #657, opened 2014-10-14
COOPR-492: Add ability to configure server to use TLS

artemmakogon:
Please review.


artemmakogon:
This build fails because lib co.cask.http:netty-http:jar:0.6.0-SNAPSHOT is absent on build machine.
I have updated netty-http library version, because I will update this library according to task COOPR-492.
@albertshau , Please advise how to fix this issue?

albertshau:
you will have to do the netty-http change first, it will need to be merged, then the snapshot will get built and pushed which will then make it available

albertshau:
but this particular change shouldn't require the updated version right?  Looks like it's just doing 1 way ssl which I think is available in whatever netty-http version was there before.

coopr-server/config/defaults/load-defaults.sh
```diff
@@ -15,11 +15,15 @@ then
   dirs="$dirs plugins/providertypes plugins/automatortypes"
 fi
 
+if [ $COOPR_SERVER_URI = "https://localhost:55054" ]; then
```
albertshau: should be checking for https, not just for https on localhost.  You could also just make CURL_PARAMETERS an env variable that can be set directly by whoever is running it.  That way it will "work" with 2 way auth later since we could set that env to include the required --cert and --cacert arg

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
```diff
@@ -53,6 +56,12 @@ private HandlerServer(Set<HttpHandler> handlers, Configuration conf) {
     builder.setBossThreadPoolSize(1);
     builder.setWorkerThreadPoolSize(numWorkerThreads);
 
+    if (enableSSL) {
+      String keyStoreFilePath = conf.get(Constants.SSL_KEYSTORE_PATH);
+      Preconditions.checkArgument(keyStoreFilePath != null);
```
albertshau: include an error message like Constants.SSL_KEYSTORE_PATH + " not specified." so that somebody reading the log immediately knows what to change.

coopr-server/src/main/java/co/cask/coopr/provisioner/mock/MockProvisionerService.java
```diff
@@ -70,9 +85,18 @@ public MockProvisionerService(String id, String serverUrl, int totalCapacity, lo
     builder.setExecThreadPoolSize(5);
     builder.setBossThreadPoolSize(1);
     builder.setWorkerThreadPoolSize(10);
-    this.httpService = builder.build();
 
-    this.httpClient = HttpClients.createDefault();
+    if (enableSSL) {
+      File keyStore = new File(Resources.getResource("cert.jks").getFile());
```
albertshau: this should be something we can set instead of hardcoded and included in the project.  

On a separate note, this will make the mock provisioner run with ssl, which will require a change in the server's HttpProvisionerRequestService to work correctly.  This is also not strictly required to add SSL to the server, and should be separated into a separate pull request.

coopr-server/src/main/java/co/cask/coopr/runtime/MockProvisionerMain.java
```diff
@@ -58,11 +58,12 @@ public void init(String[] args) {
       long msBetweenTasks = cmd.hasOption('r') ? Long.valueOf(cmd.getOptionValue('r')) : 1000;
       long taskMs = cmd.hasOption('d') ? Long.valueOf(cmd.getOptionValue('d')) : 1000;
       int failureRate = cmd.hasOption('f') ? Integer.valueOf(cmd.getOptionValue('f')) : 0;
-      String serverUrl = "http://" + host + ":" + port + Constants.API_BASE;
-      LOG.info("id = {}, capacity = {}, server url = {}, task frequency = {}, task duration = {}, failure rate = {}",
-               id, capacity, serverUrl, msBetweenTasks, taskMs, failureRate);
+      boolean enableSSL = cmd.hasOption("ssl");
```
albertshau: Not required for this pull request, but it may be time to move these into a config instead of as command line args

coopr-standalone/bin/coopr.sh
```diff
@@ -151,6 +151,23 @@ elif [ $RUBY_VERSION_MAJOR -eq 1 ] && [ $RUBY_VERSION_MINOR -lt 9 ]; then
...
+COOPR_PROTOCOL=http
+line=`awk '/server.ssl.enable/{print NR; exit}' ${COOPR_SERVER_CONF}coopr-site.xml`
+line=$((line+1))
+ssl=`sed -n "${line}p" ${COOPR_SERVER_CONF}coopr-site.xml | awk -F"<|>" '{print $3}'`
+if [ ! -z $ssl ] && [ $ssl = "true" ]; then
+  COOPR_PROTOCOL=https
+fi
+export COOPR_SERVER_URI=$COOPR_PROTOCOL://localhost:55054
+
+line=`awk '/server.disable.certificate.check/{print NR; exit}' ${COOPR_SERVER_CONF}coopr-site.xml`
```
albertshau: similar comment as the one in load_defaults.  It also looks like this setting is not in the server config so it wouldn't be in coopr-site.xml  (and it shouldn't be since this is a property of clients and not of the server).

albertshau:
we should include a unit test to make sure ssl is working.

artemmakogon:
@albertshau I have made necessary changes. Please review

albertshau:
couple small comments, then looks good to merge

artemmakogon:
@albertshau fixed

PR #655, opened 2014-10-14
Feature/tests rev1

rns4731:
Protractor tests for providers.
Minor bug fix for creating providers.

PR #660, opened 2014-10-15
protractor login helper

gaarf:
- adds login helper for integration test
- demonstrate usage with simple test for tenants section
- fix for testability of auto-refreshing ticker in tenants list

### DIFF
--- a/coopr-ngui/app/main.js
+++ b/coopr-ngui/app/main.js
@@ -89,6 +89,11 @@ angular
       container: '#alerts > .container',
       duration: 3
     });
+
+    // speed up integration tests
+    if(window.name==='PROTRACTOR') {
+      $alertProvider.defaults.duration = false;
+    }
   })
 
   .config(function ($compileProvider) {
--- a/coopr-ngui/app/main.js
+++ b/coopr-ngui/app/main.js
@@ -86,7 +86,8 @@ angular
   .config(function ($alertProvider) {
     angular.extend($alertProvider.defaults, {
       animation: 'am-fade-and-scale',
-      container: '#alerts > .container'
+      container: '#alerts > .container',
+      duration: 3
     });
   })
 
@@ -100,13 +101,13 @@ angular
     cfpLoadingBarProvider.includeSpinner = false;
   })
 
+
   .run(function ($rootScope, $alert, MYAPI_EVENT) {
     $rootScope.$on(MYAPI_EVENT.error, function (event, rejection) {
       $alert({
         title: 'API error '+rejection.status, 
         content: rejection.data || 'could not connect to the server', 
-        type: 'danger', 
-        duration: 3
+        type: 'danger'
       });
     });
   });
--- a/coopr-ngui/test/e2e/sections/tenant.js
+++ b/coopr-ngui/test/e2e/sections/tenant.js
@@ -24,36 +24,29 @@ describe('tenants section', function() {
   });
 
 
-  describe('when logged in', function() {
+  it('while logged in....', function() {
+    helper.loginAsAdmin();
 
-    beforeEach(helper.loginAsAdmin);
-    afterEach(helper.logout);
-    // beforeAll/afterAll not in Jasmine yet... but soon?
-    // cf https://github.com/angular/protractor/issues/103
+    browser.get('/#/tenants');
+
+    expect(
+      element(by.css('body')).getAttribute('class')
+    ).toContain('state-tenants-list');
 
 
-    it('/tenants renders list view', function() {
-      browser.get('/#/tenants');
 
-      expect(
-        element(by.css('body')).getAttribute('class')
-      ).toContain('state-tenants-list');
+    browser.get('/#/tenants/edit/superadmin');
 
-    });
+    expect(
+      element(by.css('body')).getAttribute('class')
+    ).toContain('state-tenants-edit');
 
-    it('/tenants/edit/superadmin renders edit view', function() {
-      browser.get('/#/tenants/edit/superadmin');
+    expect(
+      element(by.model('model.name')).getAttribute('value')
+    ).toBe('superadmin');
 
-      expect(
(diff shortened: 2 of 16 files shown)


gaarf:
relates to #655 - @nistala please have a look, we can switch to `beforeAll` when/if it ships to jasmine.

coopr-ngui/app/main.js
```diff
@@ -86,8 +86,14 @@ angular
   .config(function ($alertProvider) {
     angular.extend($alertProvider.defaults, {
       animation: 'am-fade-and-scale',
-      container: '#alerts > .container'
+      container: '#alerts > .container',
+      duration: 3
```
rns4731, line 90: what is this duration? why was it removed above

coopr-ngui/test/e2e/sections/tenant.js
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(
+      element(by.css('body')).getAttribute('class')
+    ).not.toContain('state-tenants-list');
+  });
+
+
+  describe('when logged in', function() {
+
+    beforeEach(helper.loginAsAdmin);
```
rns4731: you get the beforeAll functionality by ordering in the spec, there's no need to login and logout after each test, its just adds more time for the test to run and test things you've already tested before. I have it implemented in my PR

coopr-ngui/app/main.js
gaarf, line 90: _adding_ a default duration for the alerts instead of specifying it everywhere. Then disabling duration when running in protractor, to speed things up.

coopr-ngui/test/e2e/sections/tenant.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * e2e tests for the /tenants section
+ */
+
```
rns4731, line 4: this tests rendering of the page, do you plan on testing adding and removing tenants?

gaarf, line 4: not in this PR ;-)

coopr-ngui/test/e2e/sections/tenant.js
gaarf: yeah you can rely on the `it`s being called in order within a given file, just not throughout the whole suite. So each file has to login, and logout.

coopr-ngui/test/protractor-help.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+
+      browser.get('/#/login');
+
+      element(by.id('loginTenant')).clear().sendKeys(tenant);
+      element(by.id('loginUsername')).clear().sendKeys(username);
+      element(by.id('loginPassword')).clear().sendKeys(password);
+      element(by.partialButtonText('Submit')).click();
+
+      browser.wait(isLoggedIn, 5000);
```
rns4731, line 36: browser.waitForAngular is already called so these tests act synchronous even though they are not, you should run your tests without having to set manual waits

coopr-ngui/test/protractor-help.js
```diff
@@ -0,0 +1,69 @@
...
+        browser.wait(ddIsOpen, 5000);        
+      }
+
+      element(by.css('.dropdown-menu a[ng-click^="logout"]')).click();
+
+      browser.wait(function() {
+        return element(
+          by.cssContainingText('#alerts .alert-info', 'You are now logged out')
+        ).isPresent();
+      }, 5000);
```
rns4731, line 56: same as above.

gaarf, line 56: but these are not tests, they are the helper to login / logout. There's no `expect()`.

coopr-ngui/test/e2e/sections/tenant.js
rns4731: i've noticed the order in which you put them is the order they are run, are you seeing situations where this hasn't been true?

coopr-ngui/test/protractor-help.js
rns4731, line 56: right - you can add a simple test that checks if the page shows that you're logged in, checking the url, or some div, then you this isn't required - you're also counting on this to take a maximum of 5 seconds

coopr-ngui/test/e2e/sections/tenant.js
gaarf: not at this stage in this project, but as more tests are added, it's not uncommon to want to distribute tests across multiple machines to speed things up. The segmentation is done by suite and/or by file. Also you may want to run a single file locally. To achieve that, each test file has to be idempotent wrt to browser state and fixtures, and hence do it's own setup/teardown.

coopr-ngui/test/protractor-help.js
gaarf, line 56: but I don't want a test here.... 
screen shot 2014-10-15 at 11 42 02 am
want this number to be accurate!

coopr-ngui/test/e2e/sections/tenant.js
rns4731: e2e tests will by definition rely on ordering and sequences. Short of putting all tests into 1 file, this is a bad idea and adds a bunch on unnecessary actions which just take up time. https://github.com/angular/protractor/blob/master/lib/taskScheduler.js  - it looks like they do run in order and it looks like it chooses suites or specs giving precedence to suites. Let's not worry about problems we don't have yet, we're not using suites or distributing tests.

coopr-ngui/test/protractor-help.js
rns4731, line 56: it's accurate if you test login and logout once, which you should be doing.

coopr-ngui/test/e2e/sections/tenant.js
gaarf: I insist that we should not rely on order of files in our tests. Each file should work on it's own.

rns4731:
LGTM

PR #662, opened 2014-10-15
Angular 1.3.0 + html5mode

gaarf:
pure win :8ball:


gaarf:
WTF sauce? it works fine locally....

gaarf:
:tada: https://saucelabs.com/jobs/801d4bd8ec704d7f8fc084fd9cf011d3

coopr-ngui/app/js/directives/focus/focus.js
```diff
@@ -20,7 +20,7 @@ function myFocusDirective (myFocusManager) {
     link: function (scope, element, attrs) {
 
       attrs.$observe('myFocus', function (newVal) {
-        console.log('[myFocus] watching', newVal);
+        // console.log('[myFocus] watching', newVal);
```
ajainarayanan: Can we remove the comments? We have git to check out what the console.log statement used to be.

gaarf: done ;-)

coopr-ngui/server.js
```diff
@@ -34,51 +35,58 @@ console.log(color.hilite(pkg.name) + ' v' + pkg.version + ' starting up...');
...
+  httpIndexLogger,
+  function (req, res) {
+    res.sendFile(__dirname + '/dist/index.html');
+  }
+]);
+
+app.listen(COOPR_UI_PORT, '0.0.0.0', function () {
+  console.log(httpLabel+' listening on port %s', COOPR_UI_PORT);
+});
+
```
rns4731, line 91: :) back to express

coopr-ngui/server.js
```diff
@@ -34,51 +35,58 @@ console.log(color.hilite(pkg.name) + ' v' + pkg.version + ' starting up...');
...
+// any other path, serve index.html
+app.all('*', [
+  httpIndexLogger,
+  function (req, res) {
+    res.sendFile(__dirname + '/dist/index.html');
+  }
+]);
+
+app.listen(COOPR_UI_PORT, '0.0.0.0', function () {
+  console.log(httpLabel+' listening on port %s', COOPR_UI_PORT);
```
rns4731: console.info would be better

coopr-ngui/test/e2e/sections/tenant.js
```diff
@@ -8,11 +8,11 @@ var helper = require('../../protractor-help');
 describe('tenants section', function() {
 
   it('should redir to login when logged out', function() {
-    browser.get('/#/tenants');
+    browser.get('/tenants');
 
     expect(
       browser.getLocationAbsUrl()
-    ).toMatch(/\/#\/login/);
+    ).toMatch(/\/login$/);
```
rns4731, line 15: would these pass in browsers that don't support history api?

coopr-ngui/server.js
```diff
@@ -34,51 +35,58 @@ console.log(color.hilite(pkg.name) + ' v' + pkg.version + ' starting up...');
...
+  res.send('angular.module("'+pkg.name+'.config", [])' + 
+            '.constant("MY_CONFIG",'+data+');');
+});
+
+// serve static assets
+app.get(/\/(bundle|fonts|partials|img)\/.*/, [
+  httpStaticLogger,
+  express.static(__dirname + '/dist', {
+    index: false
+  })
```
rns4731, line 69: app.use('some static prefix', express.static(__dirname + '/dist'));

coopr-ngui/server.js
```diff
@@ -34,51 +35,58 @@ console.log(color.hilite(pkg.name) + ' v' + pkg.version + ' starting up...');
...
+    COOPR_SERVER_URI: COOPR_SERVER_URI,
+    COOPR_CORS_PORT: COOPR_CORS_PORT,
+    authorization: req.headers.authorization
+
   });
+  res.header({ 
+    'Content-Type': 'text/javascript',
+    'Cache-Control': 'no-store, must-revalidate'
+  });
+  res.send('angular.module("'+pkg.name+'.config", [])' +
```
ajainarayanan, line 60: Can we make the `PKG.name` that we use in angular a global variable so that if we need to change the name of later at some point we don't have to change the name in package.json in addition to the gulp file change when we wrap the modules is IIFE.

Just curious if it is even possible to add a variable to window object in gulp build.

coopr-ngui/test/e2e/sections/tenant.js
gaarf, line 15: yes! the regexp should match both `/login` and `/#/login` :-)

coopr-ngui/server.js
gaarf: what's the difference? (in node)

ajainarayanan:
Except for some minor comments :+1:  It looks good to me.

coopr-ngui/server.js
```diff
@@ -34,51 +35,58 @@ console.log(color.hilite(pkg.name) + ' v' + pkg.version + ' starting up...');
...
-        });
-
-        res.end(data);
-      }
-    ]
-  })
-  .listen(COOPR_UI_PORT, '0.0.0.0', function () {
-    console.log(httpLabel+' listening on port %s', COOPR_UI_PORT);
+
+var express = require('express'),
```
rns4731: best if imports are at the top of the file

coopr-ngui/server.js
gaarf, line 60: package.json is the only place you need to go to edit the project's name.

coopr-ngui/server.js
gaarf, line 69: are you commenting on the use of `app.get` or the use of a regex as pathspec? or both?

coopr-ngui/server.js
rns4731: the weight associated with it in case we choose to switch up the logging in the future, visually probably nothing.

coopr-ngui/server.js
gaarf: agreed, will fix.

coopr-ngui/server.js
rns4731, line 60: its strange that the angular init is coming from the server side, can we make an xhr request and have this just respond with JSON. Then once the /config returns use that do the angular.module on the client side

coopr-ngui/server.js
rns4731, line 69: app.use & express.static if i remember correctly will serve static assets, any reason why we want to limit it to the regex pathspec?

coopr-ngui/server.js
gaarf, line 60: we tried that, it doesnt work because then we have to call angular.bootstrap manually. Which is fine, except that protractor doesnt work if we do that instead of using `ng-app` :-(

coopr-ngui/server.js
rns4731, line 69: http://expressjs.com/api.html#express.static

gaarf, line 69: yes, to get html5 pushState mode to work, we need any path that doesnt match a static file to serve the index.html - not a 404. that's what the `app.all('*', ...` does.

coopr-ngui/server.js
rns4731, line 60: got it.

ajainarayanan, line 60: It is just the app name. I am not sure how do you mean bad here. 'Technically' angular is a global variable, but not the functionalities inside it. Thats my understanding in making this just client side

coopr-ngui/server.js
rns4731, line 69: seems counterintuitive and incorrect to return index.html for 404s, if i make a call to /static/foo.jpg, shouldn't it return a 404 if it doesn't exist, if that returns a index.html, does angular do something else to determine that it's a 404? we could revisit this if it was required for html5mode to work.

rns4731:
few comments, otherwise :+1:

coopr-ngui/server.js
rns4731, line 60: agree - don't think making the app name global is the same thing as declaring global variables in your code, by loading it from a config, its essentially doing the same thing.

coopr-ngui/server.js
gaarf, line 69: good point. Will look into having real 404s for paths under the "static" directories that don't match any static file.

PR #666, opened 2014-10-16
add missing default index.html on CentOS apache installs

duddun:


coopr-provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator/cookbooks/php_modules/recipes/default.rb
```diff
@@ -30,3 +30,12 @@
     end
   end
 end
+
+cookbook_file "{node['apache']['docroot_dir']}/index.html"
```
dereklwood: should be `"#{node...`

dereklwood:
typo, otherwise LGTM

PR #669, opened 2014-10-16
Missing do makes for many problem

wolf31o2:


wolf31o2:
Waste of time duplicate...

PR #677, opened 2014-10-17
Feature/ngui static 404

gaarf:
Show a real 404 for requests to any of the static paths that do not match a static file.


ajainarayanan:
:+1: looks good. If the last two commits are adding and reverting changes can they both be removed from this PR?

gaarf:
The code is removed, the commits cannot (or I don't know how).

My bad for pushing to wrong branch ;-)

ajainarayanan:
@gaarf No issues. You could rebased your changes in your feature branch.

`git rebase -i HEAD~4` and then removed those two commits. This would have changed the commit #s but thats fine as you can force push those to your feature (since it is your feature branch and you can do whatever you want :P)

PR #678, opened 2014-10-17
Feature/ngui doublerange fix

gaarf:
input[type=range] behavior changed with angular 1.3

this PR fixes doublerange directive, but at same time gives up on using html5 mode in coopr because that UI is not practical for the end user.


coopr-ngui/app/js/directives/doublerange/doublerange.js
```diff
@@ -20,58 +20,50 @@ function myDoublerangeDirective ($log) {
 
       scope.html5 = attrs.html5 && attrs.html5!=='false';
```
ajainarayanan, line 21: ?? Why such condition? `scope.html5 = !!scope.html5` would suffice right? if it is undefined or null it would be false otherwise true.

coopr-ngui/app/js/directives/doublerange/doublerange.js
```diff
@@ -20,58 +20,50 @@ function myDoublerangeDirective ($log) {
...
 
-        if(newVal && newVal.max) {
-          // watch out for max out!
-          if(newVal.max < oldVal.max) { 
-            if(scope.maxout) {
-              scope.maxout = false;
-            }
+      if(scope.html5) {
+
+        scope.$watch('maxout', function (newVal) {
```
ajainarayanan, line 38: To me it looks cyclic. We are watching `maxout` and `max`. When one changes we change the other.

`maxout` changes we change `max`. Now `max` watch function is executed and again `maxout` is changed?

coopr-ngui/app/partials/templates/form-tabs/constraints.html
```diff
@@ -7,7 +7,7 @@
     <my-doublerange
       data-model="model.constraints.size"
       data-min="1"
```
ajainarayanan, line 9: In the model we pass in a min and max and we have a data-min in the directive? If it is different can we have a different name?

coopr-ngui/app/js/directives/doublerange/doublerange.js
```diff
@@ -20,58 +20,50 @@ function myDoublerangeDirective ($log) {
...
+            }
           }
-        }
-      });
+        });
 
+      }
 
 
       scope.getConstraint = function (constraint, range) {
```
ajainarayanan: too damn high!

coopr-ngui/app/js/directives/doublerange/doublerange.js
gaarf, line 21: to support `data-html5="false"`, where "false" is a string.

coopr-ngui/app/js/directives/doublerange/doublerange.js
gaarf: Lol yes, will refactor. It's an exercise in how many variables we can have that include the word `max`.

coopr-ngui/app/partials/templates/form-tabs/constraints.html
gaarf, line 9: these attributes `data-min` and `data-max` define the bounds within which the doublerange should be constrained. In the JS they are bound to `scope.inputMin` and  `scope.inputMax`. What alternative naming would you suggest?

coopr-ngui/app/js/directives/doublerange/doublerange.js
```diff
@@ -20,58 +20,50 @@ function myDoublerangeDirective ($log) {
...
+              Math.ceil(scope.maxThreshold/2), 
+              scope.model.min+1
+            );
+          }
+        });
 
-      // changes going out
-      scope.$watchCollection('range', function (newVal) {
-        if(scope.model) {
-          // hack around input[type=range] getting string values
```
gaarf, line 63: a mystery is why this doesnt seem to be needed anymore. I suspect it has to do with `input[type]` attr switching from number to range on the fly? Is it a feature or a happy bug? We should avoid using html5 mode until we figure this out.

coopr-ngui/app/partials/templates/form-tabs/constraints.html
ajainarayanan, line 9: `data-input-min and data-input-max`?

coopr-ngui/app/js/directives/doublerange/doublerange.js
ajainarayanan, line 21: If we pass in a value from the scope (which is a boolean) this could be avoided. Logically every view in angular is tied to a scope (I suppose). If that is the case can we expect people using directive to do this,

``` html
<my-doublerange
  html5mode ="scope.html5mode">
</my-doublerange>
```

gaarf, line 21: it would be have to be done like this:

```
<my-doublerange
  data-model="someMinMaxObject"
  data-html5 ="{{html5mode}}"
></my-doublerange>
```

gaarf:
@ajainarayanan can I get a :+1: or is there any concern I didn't address?

coopr-ngui/app/js/directives/doublerange/doublerange.js
gaarf, line 38: refactored (I think...)

ajainarayanan:
:+1: LGTM

gaarf:
WTF saucelabs :-1:

gaarf:
FML

PR #679, opened 2014-10-17
Feature/ngui noclusters

gaarf:
fix for https://issues.cask.co/browse/COOPR-534 + hide lease renewal details for non-active clustes.


ajainarayanan:
:+1: lgtm

PR #680, opened 2014-10-17
Improve Joyent disk handling for Certified Ubuntu

wolf31o2:
- [x] Check if we're already mounted at `/data`
- [x] If not, get to unmounted state
- [x] Format, create mount point, mount


duddun:
:+1:  tested

PR #681, opened 2014-10-17
Feature/coopr templates submodule

wolf31o2:
Add `caskdata/coopr-templates` as a top-level submodule. The idea is to move template-specific code here, such as `coopr-server/config/defaults/load-defaults.sh` for reuse among multiple Coopr-related projects.


dereklwood:
:+1:

PR #685, opened 2014-10-20
Feature/ngui bugfixes

gaarf:
Fixes for:

COOPR-457
ngui: cannot open cluster in new tab (unauthorized)

COOPR-64
Confirmations should mention item being acted upon

COOPR-512
Time remaining for each some steps defaults to 45 years


.travis.yml
```diff
@@ -37,22 +37,22 @@ before_install:
...
+  - 'if [ -n "$SAUCE_ACCESS_KEY" ]; then cd ../coopr-server; fi'
+  - 'if [ -n "$SAUCE_ACCESS_KEY" ]; then java -cp target/*:src/test/resources co.cask.coopr.runtime.ServerMain; fi >/dev/null 2>&1 &'
+  - 'if [ -n "$SAUCE_ACCESS_KEY" ]; then java -cp target/*:src/test/resources co.cask.coopr.runtime.MockProvisionerMain; fi >/dev/null 2>&1 &'
   - 'cd .. && sleep 5' # give server time to start
 
 script:
   - 'cd coopr-ngui'
-  # - 'if [ -n "$SAUCE_ACCESS_KEY" ]; then npm run protractor; fi'
+  - 'if [ -n "$SAUCE_ACCESS_KEY" ]; then npm run protractor; fi'
   - 'node ./node_modules/gulp/bin/gulp.js jshint'
```
rns4731, line 58: please comment these out, these tests don't server much of a purpose now without interacting with Coopr Server - we should get sauce/selenium working first.

coopr-ngui/app/js/directives/confirmable/confirmable.js
```diff
@@ -5,28 +5,40 @@
...
 
-        if($window.confirm('Are you sure?')) {
-          confirmed();
-        }
+        modalScope = scope.$new(true);
 
-      };
+        modalScope.doConfirm = function() {
+          modal.hide();
+          scope.$eval(attrs.myConfirmable);
```
ajainarayanan, line 29: Not sure if it is safe to do `eval` on code that we don't know. Is there a better/safer way to do this?

rns4731, line 29: would really rather not do this. Why not attach to a specific attribute intended-action="delete" and then parse that out and take action based on that.

coopr-ngui/app/js/directives/confirmable/confirmable.js
```diff
@@ -5,28 +5,40 @@
...
+        };
+
+        modal = $modal({
```
rns4731, line 32: does this auto open?

coopr-ngui/server.js
```diff
@@ -63,9 +64,9 @@ app.get('/config.js', function (req, res) {
 });
 
 // serve static assets
-app.all(/\/(bundle|fonts|partials|img)\/.*/, [
+app.use('/assets', [
```
rns4731, line 67: last we discussed there was a problem with html5mode and doing this, is that resolved?

rns4731, line 67: still notice that we have an app.all that serves a index.html for all routes - that means i can type in /gibberish.png and still get a index.html

rns4731:
Would be good to know why we're using /assets now

coopr-ngui/app/js/directives/confirmable/confirmable.js
gaarf, line 29: @ajainarayanan it is code that we know, since we are writing it.

@nistala don't want to bake in that the action is $resource - related. You should be able to "confirm" anything.

coopr-ngui/server.js
gaarf, line 67: it's on purpose, that's the behavior we want for html5 mode.

gaarf:
before, we could not have any angular-run state with url starting with one of `/bundle`, `/img`, `/fonts`,`/partials`.

With this, we are just limited at not having urls starting with `/assets`.

coopr-ngui/app/js/directives/confirmable/confirmable.js
ajainarayanan, line 29: @gaarf It would be nice if we can pass in a callback function which is called when it is confirmed. 
(or)
The modal should be created in the javascript which returns a promise and when the user confirms (says yes or no) the promise should resolve stating what was done.

But here all `$eval` does is evaluate "angular" expressions and not javascript expressions. So in terms of safety it is fine but I still feel in terms of design we should try the callback or the promise way. (http://codepen.io/m-e-conroy/pen/ALsdF a reference for the idea)

gaarf, line 29: @ajainarayanan IMHO if there's a need for a more complex behavior than just "run it/dont run it", then we would be better served with a custom call to $modal. This directive is just a glorified `window.confirm()`.

gaarf:
without #655, tests pass? https://github.com/caskdata/coopr/pull/688

PR #686, opened 2014-10-20
Feature/openstack network ids

dereklwood:
- [x] openstack provider plugin support for network_ids field, for list of network uuids to associate server with


wolf31o2:
:+1: LGTM (and tested, too...)

PR #687, opened 2014-10-20
prevent potential nullpointer in server startup failure scenarios

dereklwood:
- [x] check var exists before using it... prevents an error I saw triggered when provisioner was unable to startup.


wolf31o2:
:+1: LGTM

PR #688, opened 2014-10-21
Feature/ngui bugfixes 2

gaarf:
same as #685 

COOPR-457
ngui: cannot open cluster in new tab (unauthorized)

COOPR-64
Confirmations should mention item being acted upon

COOPR-512
Time remaining for each some steps defaults to 45 years

COOPR-544
Hide theming feature


ajainarayanan:
:+1: Looks good. Sauce works and the helper/login tests have been fixed..

PR #689, opened 2014-10-21
COOPR-538 Copy pseudotty disable code from Google

wolf31o2:
More info at https://issues.cask.co/browse/COOPR-538


dereklwood:
:+1:

PR #691, opened 2014-10-21
handle deletion of invalid server

dereklwood:
- [x] Handle the case of deleting an invalid server name from Google... logs a msg and returns success.


wolf31o2:
Code :+1: LGTM... but...

The Joyent delete code's error handling looks like this:

```
      rescue NoMethodError, Fog::Compute::Joyent::Errors::NotFound
        log.warn "Could not locate server '#{providerid}'... skipping"
```

Which do we like better? Should we `debug` or `warn` on this?

dereklwood:
after some discussion, we decided INFO is the way to go.  opened https://issues.cask.co/browse/COOPR-539 to consistently implement some INFO logging.  merging this for now.

PR #692, opened 2014-10-22
Use coopr-templates submodule

wolf31o2:
Now that we have the caskdata/coopr-templates repository and it is added as a submodule, we want to use the submodule instead of any other templates. This allows us to only need to update one location for templates, while allowing contributors for templates to not get bogged down in the rest of the Coopr code.

Tested in standalone using both the Ruby and Dummy provisioners.


duddun:
:+1:

PR #694, opened 2014-10-22
Use public, private, then first address for bootstrap_ip

wolf31o2:
This resolves an issue where the provisioner is on the same network as the instance being provisioned, a floating IP address has not been requested, and there is no network named `public` or `private` to detect.


nitinmotgi:
:+1:  LGTM

PR #695, opened 2014-10-22
Feature/data uploader usage

dereklwood:
fixes COOPR-466, better handling of data-uploader usage:
- [x] print out usage info when no options given, or invalid options given, or invalid args given
- [x] exits non-zero as appropriate


wolf31o2:
:+1: LGTM

PR #697, opened 2014-10-22
Added ability to specify non default Gson object to RestClient.

amakogoncv:
So far as exists ability to create `Gson` object not only with default constructor, but also using `GsonBuilder`, I added new constructor to the `RestClient` which allows to specify non default `Gson` object for json serialization and deserialization. @albertshau , please review this changes.


albertshau:
lgtm

PR #698, opened 2014-10-22
Rest client integration tests

amakogoncv:
This PR could be merged only after https://github.com/caskdata/coopr/pull/697


albertshau:
There was no deserialize method written for ClusterDetails because serialization is being done in a custom way and it wasn't needed at the time.  For historical reasons, the json is some weird combination of various objects and is being pieced together because we didn't want to change the api at the time.

Long story short, you will have to add a deserialize method to the ClusterDetailsCodec to get it to work and it may be a little messy.

coopr-integration-testing/src/test/java/co/cask/coopr/test/client/rest/AdminRestClientTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+  @Before
+  public void setUp() {
+    adminClient = adminClientManager.getAdminClient();
+  }
+
+  @Test
+  public void testGetAllClusterTemplates() throws IOException {
+    List<ClusterTemplate> result = adminClient.getAllClusterTemplates();
+    assertTrue(result.contains(REACTOR_CLUSTER_TEMPLATE));
```
albertshau: to be consistent with the other tests, you can just stick with Assert.assertTrue instead of importing the static methods to call assertTrue

coopr-integration-testing/src/test/java/co/cask/coopr/test/client/rest/AdminRestClientTest.java
```diff
@@ -0,0 +1,217 @@
...
+    assertTrue(result.contains(HDFS_CLUSTER_TEMPLATE));
+    assertTrue(result.contains(HADOOP_DISTRIBUTED_CLUSTER_TEMPLATE));
+    assertTrue(result.size() == 3);
```
albertshau: Assert.assertEquals(3, result.size()) will give a more descriptive error message if it fails

albertshau: similar feedback for other places in the test that this is done

coopr-integration-testing/src/test/java/co/cask/coopr/test/client/rest/ClusterRestClientTest.java
```diff
@@ -0,0 +1,253 @@
...
+  @Test
+  public void testCreateCluster() throws IOException {
+    ClusterCreateRequest createRequest = ClusterCreateRequest.builder()
+      .setName(EXPECTED_NEW_CLUSTER_NAME)
+      .setClusterTemplateName(Entities.ClusterTemplateExample.REACTOR.getName())
+      .setNumMachines(10)
+      .setDNSSuffix("ro.test.com")
+      .build();
+    String newClusterId = clusterClient.createCluster(createRequest);
+    assertFalse(Strings.isNullOrEmpty(newClusterId));
```
albertshau: this should be null right?  better to Assert.assertNull.

amakogoncv: No, this should not be null, on the contrary. But it looks quite unclear, you are right. Thanks

albertshau:
lgtm, travis failures are unrelated and should be fixed upon merge.

PR #699, opened 2014-10-22
Reverse order to prevent null pointer

wolf31o2:
This is already done in `chef_solo_automator` and `fog_provider` but we missed changing it in `shell_automator` to match.


dereklwood:
:+1:

PR #703, opened 2014-10-22
Better output on credential failures

wolf31o2:
The important takeaway is assigning an error message to `@result['stderr']` which makes it available to callbacks.


dereklwood:
:+1:

PR #704, opened 2014-10-23
change load defaults script location based on change from backend

rns4731:
Changes in tests for defaults being moved out from backend and fixing travis config for change in location.

### DIFF
--- a/coopr-e2e/test/e2e/providers/provider-spec.js
+++ b/coopr-e2e/test/e2e/providers/provider-spec.js
@@ -41,7 +41,8 @@ describe('providers test', function () {
     expect(zone).toEqual('5');
   });
 
-  it('should create a provider', function () {
+  it('should create, store and delete a provider', function () {
+    // Create
     element(by.name('name')).sendKeys('Testprovider');
     element(by.css('#inputProviderDescription')).sendKeys('Test description');
     element(by.cssContainingText('option', 'google')).click();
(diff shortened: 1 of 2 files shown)


gaarf:
Questions, maybe @albertshau or @wolf31o2 can chime in.
- After running `coopr-server/templates/mock/load-mock.sh`, is there a way to reset the data other than recompiling? If one runs `load-mock.sh` several times, will there be duplicates?
- it would be nice to have something like a `npm run loadmocks` command, in either/or/both `coopr-ngui/package.json` & `coopr-e2e/package.json`
- @nistala please edit `coopr-ngui/README.md` with updated info re: how to load the mocks and run the tests locally.

coopr-e2e/test/e2e/providers/provider-spec.js
```diff
@@ -52,78 +53,40 @@ describe('providers test', function () {
...
     providersList = element.all(by.repeater('item in list'));
-    expect(providersList.count()).toEqual(7);
-  });
-
-  it('should navigate to provider edit page and show fields.', function () {
-    browser.get('/providers/edit/google');
+    expect(providersList.count()).toEqual(1);
+    var providerText = providersList.first().getText();
+    expect(providerText).toMatch('Testprovider Test description Edit\nDelete');
+    providersList.first().element(by.css('.edit-btn')).click();
```
gaarf, line 66: please remove the `edit-btn` class in the template, and use `by.cssContainingText('.btn', 'Edit')` here.

coopr-e2e/test/e2e/providers/provider-spec.js
```diff
@@ -52,78 +53,40 @@ describe('providers test', function () {
...
-  //   expect(formfields.get(4).element(by.css('input')).getAttribute('value')).toBe('sshfoo');
-  //   expect(formfields.get(5).element(by.css('input')).getAttribute('value')).toBe('resourcefoo');
-
-  //   // !TODO call /reset to get default data back to normal once default data has been implemented.
-  //   // browser.get('/reset');
-  // });
+    // Delete
+    browser.get('/providers');
+    providersList = element.all(by.repeater('item in list'));
+    providersList.first().element(by.css('.delete-btn')).click();
```
gaarf, line 85: please remove the `delete-btn` class in the template, and use `by.cssContainingText('.btn', 'Delete')` here.

wolf31o2:
@gaarf re: reset... there's no data loaded by us. The new `load-mock.sh` only loads the necessary plugin support data. Everything else is empty. The only way to reset the data is to shut down the server, delete the data directory, and restart the server. Something like a `/reset` endpoint is only viable with the built-in Derby database. It wouldn't be possible using any other database, as management is done by the database administrator, not the server process. Database management is done outside Coopr, for the most part.

gaarf:
@wolf31o2 thanks. IMHO resetting mocked data really only matters for development, running tests locally. I think it would be nice to have a shell script to do this (shut down the server, delete the data directory, and restart the server).

gaarf:
merging this now else tests won't pass for anybody. @nistala please address documentation and tooling remarks in a future PR.

wolf31o2:
@gaarf file a JIRA for the helper script and we'll get it added... :smile:

gaarf:
@wolf31o2 done :-) https://issues.cask.co/browse/COOPR-547

PR #705, opened 2014-10-23
COOPR-520: Add mutual auth support

artemmakogon:
Please review


wolf31o2:
This needs some slight changes for the recent move of templates to caskdata/coopr-templates and the load-defaults.sh script, along with it. The script is now just `bin/load-templates.sh` in the `caskdata/coopr-templates` repository. Once that has been updated and merged, the reference to the submodule should be updated in this branch to point to the new ref...

```
git submodule init
git submodule update
git add coopr-templates # once it's updated and merged
git commit -m "Update coopr-templates ref for COOPR-520"
```

coopr-server/src/test/java/co/cask/coopr/http/MutualAuthServerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    Future<HttpResponse> future = Executors.newSingleThreadExecutor().submit(new Callable<HttpResponse>() {
+      public HttpResponse call() throws Exception {
+        return doGet(String.format("http://%s:%d/status", HOSTNAME,
+                                   handlerServer.getBindAddress().getPort()));
+      }
+    });
+    future.get(5000, TimeUnit.MILLISECONDS);
+  }
+
+  public static HttpResponse doGet(String url) throws Exception {
```
albertshau: this is a little confusing, thought the doGet calls elsewhere in the test were using the parent's doGet until I saw this.  Maybe call this something else like doSecureGet

albertshau: can this be private

coopr-server/src/test/java/co/cask/coopr/http/HttpsServerTest.java
```diff
@@ -79,7 +79,7 @@ public HttpResponse call() throws Exception {
 
   public static HttpResponse doGet(String url) throws Exception {
```
albertshau: can this be private

coopr-standalone/bin/coopr.sh
```diff
@@ -155,12 +155,24 @@ fi
...
+cert_line=`awk '/server.ssl.trust.cert.path/{print NR; exit}' ${COOPR_SERVER_CONF}coopr-security.xml`
+cert_line=$((cert_line+1))
+export TRUST_CERT_PATH=`sed -n "${cert_line}p" ${COOPR_SERVER_CONF}coopr-security.xml | awk -F"<|>" '{print $3}'`
+
+cert_pass_line=`awk '/server.ssl.trust.cert.password/{print NR; exit}' ${COOPR_SERVER_CONF}coopr-security.xml`
+cert_pass_line=$((cert_pass_line+1))
+export TRUST_CERT_PASSWORD=`sed -n "${cert_pass_line}p" ${COOPR_SERVER_CONF}coopr-security.xml | awk -F"<|>" '{print $3}'`
+
+if [ ! -z TRUST_CERT_PATH ] && [ ! -z TRUST_CERT_PASsWORD ]; then
+  export CERT_PARAMETER="--cert ${TRUST_CERT_PATH}:${TRUST_CERT_PASSWORD}"
```
albertshau, line 147: does this work? I thought curl doesn't handle java keystore formats?

albertshau, line 147: If not it may be better to have the cert and optional password as env settings that are set directly by the user instead of tying it to what the server is using.

albertshau:
for this, let's just concentrate on getting it so that we can configure the server to use mutual auth and not worry about other components, but do them separately.

coopr-standalone/bin/coopr.sh
wolf31o2, line 147: That does not work. Neither does the Ruby.

albertshau:
I think there needs to be a high level discussion on the security story and how we're planning on implementing it before we go too far into the provisioner and ui parts.

albertshau:
lets leave out the RestHelper provisioner changes and do those separately after a discussion with @wolf31o2 and @dereklwood on how they want to handle ssl in the provisioner.

albertshau:
otherwise lgtm

coopr-server/templates/mock/load-mock.sh
```diff
@@ -19,7 +19,7 @@ for d in ${dirs} ; do
       --header "Coopr-ApiKey:${COOPR_API_KEY}" \
       --header "Coopr-TenantID:${COOPR_TENANT}" \
       --connect-timeout ${TIMEOUT} --data @${f} \
-      ${COOPR_SERVER_URI}/v2/${d}/${f/.json/}
+      ${COOPR_SERVER_URI}/v2/${d}/${f/.json/} $CERT_PARAMETER
```
wolf31o2: Braces around variables, please...

coopr-standalone/bin/coopr.sh
```diff
@@ -137,12 +137,16 @@ fi
...
+export COOPR_SSL=`sed -n "${line}p" ${COOPR_SERVER_CONF}/coopr-site.xml | awk -F"<|>" '{print $3}'`
+if [ ! -z ${COOPR_SSL} ] && [ ${COOPR_SSL} = "true" ]; then
   COOPR_PROTOCOL=https
 fi
 export COOPR_SERVER_URI=${COOPR_PROTOCOL}://localhost:55054
 
+if [ ! -z ${TRUST_CERT_PATH} ] && [ ! -z ${TRUST_CERT_PASSWORD} ]; then
+  export CERT_PARAMETER="--cert ${TRUST_CERT_PATH}:${TRUST_CERT_PASSWORD}"
+fi
+
```
wolf31o2, line 149: This does not work. There is no support in curl for Java keystores.

albertshau, line 149: there's nothing specific to java keystores here.  I think this is fine for now, but we will want to revisit how to handle ssl in our scripts.

artemmakogon, line 149: @wolf31o2 we have discussed this part with Albert. TRUST_CERT_PATH is not a path to '.jks' file.

albertshau:
lgtm

PR #706, opened 2014-10-23
Update coopr-templates reference to latest master

wolf31o2:


wolf31o2:
OTS approved by @duddun

PR #710, opened 2014-10-24
Update bash scripts to work in --posix mode

wolf31o2:
This cleans up the scripts to be a consistent coding style, removes unused code, does some optimizations, and generally makes things more standardized.


duddun:
:+1:

PR #712, opened 2014-10-24
revert to angular 1.2.x + other fixes missed in previous PRs

gaarf:
I think we are being hit by the issues mentioned on https://docs.angularjs.org/api/ng/directive/select

```
 ngModel compares by reference, not value. This is important when binding to an array of objects. See an example in http://jsfiddle.net/qWzTb/.
```


ajainarayanan:
:+1: LGTM.. Fixes Sauce for develop

gaarf:
hope so....

PR #716, opened 2014-10-27
COOPR-550: Admin and Cluster CLI commands integration tests

ghost:


coopr-integration-testing/src/test/java/co/cask/coopr/test/cli/AdminCommandsTest.java
```diff
@@ -0,0 +1,192 @@
...
+    execute(command);
+    checkError();
+  }
+
+  @Test
+  public void testDeleteProvider() throws InvalidCommandException, UnsupportedEncodingException {
+    String getCommand = String.format(Constants.GET_PROVIDER_COMMAND, Entities.JOYENT);
+    String deleteCommand = String.format(Constants.DELETE_PROVIDER_COMMAND, Entities.JOYENT);
+
+    execute(getCommand);
```
albertshau: The delete tests are very similar, you could refactor into one common private method that takes a get command, expected result of the get, and a delete command.

coopr-integration-testing/src/test/java/co/cask/coopr/test/cli/AdminCommandsTest.java
```diff
@@ -0,0 +1,192 @@
...
+  @BeforeClass
+  public static void beforeClass()
+    throws URISyntaxException, NoSuchFieldException, IllegalAccessException, IOException {
+    createCli(ADMIN_ACCOUNT);
+  }
+
+  @Test
+  public void testListTemplates() throws InvalidCommandException, UnsupportedEncodingException {
+    execute(Constants.LIST_TEMPLATES_COMMAND);
+    checkCommandOutput(Arrays.asList(HADOOP_DISTRIBUTED_CLUSTER_TEMPLATE, REACTOR_CLUSTER_TEMPLATE,
```
albertshau: it may be the case that these are always returned in the same order, but there is nothing in the api that guarantees this.  It would be safer to convert the output into a set, then compare sets.

coopr-integration-testing/src/test/java/co/cask/coopr/test/cli/ClusterCommandsTest.java
```diff
@@ -0,0 +1,277 @@
...
+  @BeforeClass
+  public static void beforeClass()
+    throws URISyntaxException, NoSuchFieldException, IllegalAccessException, IOException {
+    createCli(ADMIN_ACCOUNT);
+  }
+
+  @Test
+  public void testListClusters() throws InvalidCommandException, UnsupportedEncodingException {
+    execute(Constants.LIST_CLUSTERS_COMMAND);
+    List<ClusterSummary> result = getListFromOutput(new TypeToken<List<ClusterSummary>>() {}.getType());
```
albertshau: you could get a set back which would make the comparison much easier.

coopr-integration-testing/src/test/java/co/cask/coopr/test/cli/ClusterCommandsTest.java
```diff
@@ -0,0 +1,277 @@
...
+    execute(command);
+    ClusterDetails result = getObjectFromOutput(ClusterDetails.class);
+    Assert.assertNotNull(result);
+    Assert.assertEquals(ClusterAction.CLUSTER_CREATE, result.getProgress().getAction());
+    for (Node node : result.getNodes()) {
+      Assert.assertTrue(FIRST_TEST_CLUSTER.getNodeIDs().contains(node.getId()));
+    }
+    // Node ids get overwritten by the full node objects in the result object
+    Cluster expectedResult = FIRST_TEST_CLUSTER;
+    expectedResult.setNodes(ImmutableSet.<String>of());
```
albertshau: this seems wrong, probably something missing from ClusterDetailsCodec's deserialize method if the node ids are not getting populated.

albertshau: actually, getObjectFromOutput uses a gson instance without any of our codecs, so any object that has custom deserialization logic (like ClusterDetails) will not deserialize correctly.  Is the CLI also not using our codecs to output?  If not that needs to be fixed, otherwise the CLI will output something different than the rest APIs, and you won't be able to deserialize correctly here.

albertshau:
There may be an issue that needs to be fixed before this can be done.  If the CLI is not outputting with our codecs, any custom serialization logic will not take place and the output will not match the rest api output.  It will also mean that whatever is output cannot correctly be deserialized into its corresponding java object, which causes problems in these integration tests.

All the current serialization/deserialization stuff is too complicated and should be cleaned up at some point, but I'll open a separate jira for that.

ghost:
Hello, I have fixed comments above, but CLI output method still uses Gson without Coopr codecs. I can't use Coopr Gson, because output must be pretty: (GsonBuilder().setPrettyPrinting().disableHtmlEscaping()). And if I output objects with Coopr Gson, then I got one long ugly json line.

coopr-integration-testing/src/test/java/co/cask/coopr/test/cli/ClusterCommandsTest.java
albertshau: taking a look at the ClusterDetailsCodec, seems like the deserialize method needs to do a little more and calculate the node ids and call `cluster.setNodes(nodeIds)`.  Then you should be able to compare them directly instead of modifying.

albertshau:
One last comment about ClusterDetails, which wasn't fully working before your PR but should be fixed.

ghost:
Maybe I do not understand you correctly, but I need fix ClusterDetails logic? For now it is doesn't work.

albertshau:
ClusterDetails is not getting deserialized correctly, which is why the cluster you get back does not have any node ids even though it has nodes.  This needs to be fixed in the codec instead of working around the bug in the test.

ghost:
Ok, thanks. I understand now

albertshau:
lgtm pending travis tests

PR #720, opened 2014-10-27
Update to correct PATH

wolf31o2:
The start script still points to ui instead of ngui. This was working in test installs because we were installing both the old and new UI on the test machine, and the script would execute the old UI's embedded node.


wolf31o2:
:+1: OTS @gaarf

PR #722, opened 2014-10-27
Update ref after merging caskdata/coopr-templates#9

wolf31o2:
This keeps us from filling up root disks on providers with small root disks.


albertshau:
:+1:

PR #727, opened 2014-10-28
Move resources to resources directory

wolf31o2:
It'll be easier to evaluate this if you look at the individual commits. Ignore the first commit, as it's a move of all of the cookbooks, which is the main cause for the size of this PR.


duddun:
:+1:

PR #728, opened 2014-10-28
Create node_modules cookbook

wolf31o2:
This is a simple cookbook to use the `nodejs_npm` LWRP provided by the `nodejs` cookbook to allow us to install modules for Node.JS into a cluster created by Coopr.


duddun:
:+1:

PR #733, opened 2014-10-29
Feature/openstack userdata resource

dereklwood:
- [x] adds support for the Openstack user data feature in the fog provider plugin.  defines a user_data resource, so one can just upload a script to coopr as a named resource, and then configure it in their openstack coopr provider.
- [x] updates dummy provisioner


duddun:
:+1:

PR #734, opened 2014-10-29
Feature/ngui pausedeploy

gaarf:
https://issues.cask.co/browse/COOPR-555 + other minor tweaks

### DIFF
--- a/coopr-ngui/app/js/controllers/cluster/detail-ctrl.js
+++ b/coopr-ngui/app/js/controllers/cluster/detail-ctrl.js
@@ -124,6 +124,17 @@ function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, m
   $scope.doActionsModal = doActionsModal;
 
 
+
+
+  $scope.doPauseDeploy = function() {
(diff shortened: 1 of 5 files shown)


gaarf:
screen shot 2014-10-29 at 1 02 53 pm

coopr-ngui/app/js/controllers/cluster/detail-ctrl.js
```diff
@@ -43,7 +43,7 @@ function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, m
     $scope.additionalServices = [];
 
     if(data.status === 'pending') {
-      timeoutPromise = $timeout(update, 1000);
+      timeoutPromise = $timeout(update, 10000);
```
ajainarayanan, line 46: Any specific reason why the increase in the timeout? Seems to be changed uniformly in the cluster feature.

gaarf, line 46: yes, talked to @wolf31o2 , in real conditions no task would take less than 30seconds. So updating every 10seconds seems like enough.

ajainarayanan:
:+1: LGTM.

PR #735, opened 2014-10-30
Update ref to latest coopr-templates

wolf31o2:
Changes: https://github.com/caskdata/coopr-templates/compare/8080e84...3b30967


duddun:
:+1:

PR #738, opened 2014-10-30
add Digital Ocean provider plugin

duddun:


coopr-provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fog",
   "description": "Multi-cloud fog-based provider (AWS, Google, Openstack, Rackspace, Joyent)",
-  "providertypes": [ "aws", "google", "joyent", "openstack", "rackspace" ],
+  "providertypes": [ "aws", "google", "joyent", "openstack", "rackspace", "digitalocean" ],
```
wolf31o2: Alphabetize...

coopr-provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -453,5 +453,62 @@
...
+            "type": "password",
+            "tip": "Your DigitalOcean API key"
+          },
+          "digitalocean_region": {
+            "label": "Region",
+            "type": "select",
+            "options": [ "3", "4", "5", "6", "7", "8", "9" ],
+            "default": "3",
+            "override": true,
+            "tip": "DigitalOcean Region: 3=sfo1 4=nyc2 5=ams2 6=sgp1 7=lon1 8=nyc3 9=ams3"
```
wolf31o2: Since this will only display the numbers to the user, we should switch to having the options be `sfo1`, `nyc2`, etc. and do translation in the plugin code to the correct numerical value.

coopr-provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -453,5 +453,62 @@
...
+            "permissions": "0400"
+        }
+    },
+    "parameters": {
+      "admin": {
+        "fields": {
+          "api_user": {
+            "label": "Client ID",
+            "type": "text",
+            "tip": "Your DigitalOcean ClientID"
```
wolf31o2: Client ID

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider.rb
```diff
@@ -21,6 +21,7 @@
 require_relative 'fog_provider/joyent'
 require_relative 'fog_provider/openstack'
 require_relative 'fog_provider/rackspace'
+require_relative 'fog_provider/digitalocean'
```
wolf31o2: Alphabetize

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/digitalocean.rb
```diff
@@ -0,0 +1,191 @@
...
+  include FogProvider
+
+  # plugin defined resources
+  @ssh_key_dir = 'ssh_keys'
+  class << self
+    attr_accessor :ssh_key_dir
+  end
+
+  def create(inputmap)
+    flavor = inputmap['sizes']
```
wolf31o2: This is definitely going to be `inputmap['flavor']` as it comes from the worker. This is not provider-specific.

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/digitalocean.rb
```diff
@@ -0,0 +1,191 @@
...
+      # Create the server (droplet)
+      log.debug "Creating #{hostname} on DigitalOcean using flavor: #{flavor}, image: #{image}"
+      log.debug 'Invoking server create'
+      begin
+        server = connection.servers.create(
+          name: hostname,
+          image_id: image,
+          flavor_id: flavor,
+          region_id: region,
+          key_name: @ssh_keypair
```
wolf31o2: We will need to verify this. They may not accept a key name, since everything else in their API is based on integer IDs. We may need to do a lookup on the key to get its ID.

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/digitalocean.rb
```diff
@@ -0,0 +1,191 @@
...
+        )
+        server.persisted? || server.save
```
wolf31o2: This line came from Rackspace and may not be valid. In the case of OpenStack, doing a server.save attempts to create a clone of the server object and fails. I would try commenting it and see if this works.

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/digitalocean.rb
```diff
@@ -0,0 +1,191 @@
...
+      # Our fields are fog symbols
+      fields.each do |k, v|
+        instance_variable_set('@' + k, v)
+      end
+      # Confirm server
+      log.debug "Invoking server confirm for id: #{providerid}"
+      server = connection.servers.get(providerid)
+
+      # Wait until the server is ready
+      fail "Server #{server.name} is in ERROR state" if server.state == 'ERROR'
```
wolf31o2: It looks like DigitalOcean uses lowercase for these. Probably a good idea to make sure we're checking the correct case.... something like this:

```
fail "Server #{server.name} is in ERROR state" if server.state.downcase == 'error'
```

coopr-provisioner/worker/plugins/providers/fog_provider/lib/fog_provider/digitalocean.rb
wolf31o2: So... they don't use key_name... they do keys by id...

```
ssh_key_ids: [ id1, id2, id3 ]
```

This means you're going to have to do the following:

```
key_id = compute.ssh_keys.select { |k| k.name == @ssh_keypair }.first.id
```

This will give you a key_id that you can pass with ssh_key_ids to your servers.create

duddun:
Please review this when possible.

coopr-provisioner/worker/plugins/providers/fog_provider/fog_provider.json
```diff
@@ -141,6 +141,59 @@
...
+          "api_password": {
+            "label": "API key",
+            "type": "password",
+            "tip": "Your DigitalOcean API key"
+          },
+          "digitalocean_region": {
+            "label": "Region",
+            "type": "select",
+            "options": [ "sfo1", "nyc2", "ams2", "sgp1", "lon1", "nyc3", "ams3" ],
+            "default": "sfo1",
```
wolf31o2: I would change the default to `nyc2` since that appears to be the DigitalOcean default.

duddun: Good catch.  Will do.

David.

coopr-provisioner/master/conf/provisioner-default.xml
```diff
@@ -63,7 +63,7 @@
   </property>
   <property>
     <name>provisioner.log.level</name>
-    <value>info</value>
+    <value>debug</value>
```
wolf31o2: Revert this change.

wolf31o2:
One comment... otherwise, :+1: LGTM...

PR #739, opened 2014-10-30
Feature/ngui refactor part2

gaarf:
- moves app/css to app/styles and puts img under that
- split up app/routes.js and distribute among features

cf https://issues.cask.co/browse/COOPR-559


gaarf:
part 3 will come in over the week-end.

ajainarayanan:
:+1: LGTM.

PR #740, opened 2014-10-31
Added dependency to the co.cask.common project.

amakogoncv:
Common classes was taken out from the Cdap project to the separate Common repo and could be reused now in the Coopr Rest Client, for avoid code duplication. This PR contains refactor of the Java Rest Client code.
@albertshau , please, take a look.


albertshau:
:+1:

PR #744, opened 2014-11-01
Feature/ngui refactor part3

gaarf:
final step of https://issues.cask.co/browse/COOPR-559
- each feature is a module
- renames the _.directives_ module to _.commons_
- whitespace cleanup by editorconfig.org plugin on those files that were edited.


ajainarayanan:
@gaarf  Awesome!

Some minor comments,
- We discussed a hybrid structure where each feature will have its own folders of controllers(js), styles, partials (directives and services if it makes sense to put it under that feature)
- Can we re-name `_<filename>` pattern? This might be just my thinking but `_` or `__` in javascript means it is private to the class (or object?) and is not supposed to be used. Can we just name it plain names without `_`
- Technically it is not directives alone that are put under `common` right? All `directives, services and filters` are under common right? Services and Filters are added as dependency so if I use `.commons` module then it comes with it. Can we have either everything defined as dependency or everything under `.commons` module? 
  The thing that I am trying to understand is directives are added as directives under .commons module but services and filters are built as separate modules under .services and .filters and added as dependency. Any specific reason for this?

This might be trivial but might be important on a longer run but there are some formatting that we need to address. It doesn't have to be in this PR but in some future PR would be nice.
- JSON objects should not be single line (`{ "asd": "asdasd"}`)
- One space between operator and the keyword,

``` javascript
for (var i = 0; i < 10; i++) {
    ...
}

if (someCondition) {
    ...
}

function someName (args...) {
    ...
}

angular.module("someName")
       .controller("someController", function someController () {
           ...
        });
```

These are some instances but we need to create a style guide to comprehensively go through it.

ajainarayanan:
Also commit messages. :smile: 

As we discussed lets try this pattern for sometime, if doesn't work then we can switch to something else,

[Adds/Removes/Updates/Rewrites] [feature] [Brief reason why it is done]

gaarf:
:+1: will be careful re: commit messages

gaarf:
I'll remove the underscores from the module/routes files and crud/home features folders.

ajainarayanan:
:+1: Good to go!

As per our discussion,
1. If it makes sense for features (meaning if it scales for more than a couple of controllers) then we create a new folder under the feature (say controllers) and put our controller files under it. This will hold for partials(or templates) and styles (or less/css) folders as well. (not in coopr but probably in cdap)
2. Code linting will be editorconfig + jslint (or jshint)
3. We are renaming `_` from file names.
4. :+1: on making the `commons` module the way it is. It makes sense as directives depend on filters and services.

gaarf:
actually removing the underscore from file names breaks a lot of things... seems like the file order matters now :-(

gaarf:
@ajainarayanan thanks for you patience :-)

ajainarayanan:
:+1: No worries! All good :)

PR #749, opened 2014-11-04
COOPR-584 various bug fixes to standalone. Config was broken,

albertshau:
dummy server script had a typo, and the standalone script would
try to load mocks on every command.


duddun:
:+1:

PR #750, opened 2014-11-04
integrate with browserstack and travis

rns4731:
- Browserstack integration
- hardware types tets
- image types tests
- fix bug with POST request for creating image types


coopr-e2e/test/protractor-help.js
```diff
@@ -21,68 +19,20 @@ module.exports = {
...
-      if(needLogout) {
-        logout();
-      }
-
-      browser.get('/login');
-      browser.waitForAngular();
+  browser.get('/login');
+  element(by.id('loginTenant')).clear().sendKeys('superadmin');
+  element(by.id('loginUsername')).clear().sendKeys('admin');
+  element(by.id('loginPassword')).clear().sendKeys('admin');
```
gaarf: to support the future case where we want to login as something else than superadmin, these 3 statements should use the `tenant, username, password` variables, which are passed as arguments to the login() function.

coopr-e2e/test/e2e/hardwares/hardware-spec.js
```diff
@@ -6,9 +6,57 @@ var helper = require('../../protractor-help');
 'use strict';
 
 describe('hardware types test', function () {
+
+  var hardwareTypes;
+
+  it('should log in', function () {
+    helper.loginAsAdmin();
```
ajainarayanan, line 13: We have the helper login as admin. I feel if we assert that we have logged in then it would make the testcase complete. It would be redundant across all features but I feel for the sake of sticking to the "testing pattern in javascript" if the test case asserts something then it would be a valid pass. Any thoughts?

gaarf, line 13: until we have `beforeAll` and `afterAll`, I think we're kinda stuck with putting the login/logout in `it` statements. There is an `expect()` call in the helper now, so strictly speaking we are still testing _something_ :joy_cat:

rns4731, line 13: +1 helper login does have an expect in it

gaarf:
if our browserstack account is now set up properly, this LGTM :+1:

rns4731:
Addressed comments above.

PR #752, opened 2014-11-05
Feature/ngui shared modules

gaarf:
- should include cask-angular-\* bower modules in lib.js / lib.css
- update to angular 1.3.1
- fix 1.3.x select directive issue by using _track by_

### DIFF
--- a/coopr-ngui/app/main.js
+++ b/coopr-ngui/app/main.js
@@ -95,11 +95,6 @@ angular
       container: '#alerts > .container',
       duration: 3
     });
-
-    // speed up integration tests
-    if(window.name==='PROTRACTOR') {
-      $alertProvider.defaults.duration = false;
-    }
   })
 
   .config(function ($compileProvider) {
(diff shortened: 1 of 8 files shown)


ajainarayanan:
:+1: LGTM. Simple changes.

I am guessing the travis build failure is because of backend timeout.

coopr-ngui/app/main.js
```diff
@@ -95,11 +95,6 @@ angular
       container: '#alerts > .container',
       duration: 3
     });
-
-    // speed up integration tests
-    if(window.name==='PROTRACTOR') {
```
ajainarayanan, line 100: I am guessing this was causing issue in sauce and that since we moved out of sauce we removed this?

gaarf, line 100: yes it was an attempt to "speed up" sauce. we can revisit the strategy if browserstack were to become unbearably slow.

PR #754, opened 2014-11-06
checkstyle was configured incorrectly. Fixed the pom to actually

albertshau:
use the correct checkstyle file and also updated the checkstyle
file and fixed all errors and warnings.


wolf31o2:
:+1: LGTM... build is failing, though... at least one thing timed out, but it looked like there was another failure, so I didn't restart the build.

PR #755, opened 2014-11-07
Hotfix/missing dev packages

wolf31o2:
Building `fog` requires `nokogiri`, which builds `libxml2`, which is statically-linked against `libz.a`... whew!


albertshau:
lgtm but travis does not like

wolf31o2:
I re-ran the test and Travis says it's all good. Merging...

PR #760, opened 2014-11-11
[COOPR-588] Mac OS support improved for coopr-standalone/bin/coopr.sh

kpovetkin:
According to curl usage there was a discussion: https://github.com/caskdata/coopr/pull/705#discussion_r19500285


wolf31o2:
Thanks!

PR #762, opened 2014-11-11
Feature/auth service integration

amakogoncv:
Integration of COOPR server and External Authentication service from CDAP project. As was discussed before with @sreevatsanraman , I used dependency to the cdap repo, but in future it should be replaced with dependency to the common repo. 
Also, I copied class `SecurityAuthenticationHttpHandler` from the `cdap-gateway` package. This class also could be reused, so @sreevatsanraman , could you please move this class to the common auth service too? Then we could delete it from this repo. 
@albertshau , please take to review this PR.


amakogoncv:
I will fix unit tests in the next commit. Please, review code for now, to understand, is it ok for you.

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
```diff
@@ -36,15 +45,24 @@
...
+  private HandlerServer(Set<HttpHandler> handlers, Configuration conf, final CConfiguration cConf,
+                        final TokenValidator tokenValidator,
+                        final AccessTokenTransformer accessTokenTransformer,
+                        final DiscoveryServiceClient discoveryServiceClient) {
     String host = conf.get(Constants.HOST);
     int port = conf.getInt(Constants.PORT);
     int numExecThreads = conf.getInt(Constants.NETTY_EXEC_NUM_THREADS);
     int numWorkerThreads = conf.getInt(Constants.NETTY_WORKER_NUM_THREADS);
     boolean enableSSL = conf.getBoolean(Constants.ENABLE_SSL);
+    final boolean securityEnabled = conf.getBoolean(co.cask.cdap.common.conf.Constants.Security.CFG_SECURITY_ENABLED);
```
albertshau, line 64: i think names of config settings should be specific to coopr, can create coopr versions of these settings.

coopr-server/src/main/java/co/cask/coopr/runtime/ServerMain.java
```diff
@@ -131,6 +141,10 @@ public void start() {
                                       .setDaemon(true)
                                       .build()));
 
+    CConfiguration cConfiguration = CConfiguration.create();
```
albertshau, line 144: Will this try and load the cdap config? Ideally configuration class would get moved to the common module as well.

coopr-standalone/bin/coopr.sh
```diff
@@ -21,6 +21,7 @@ export COOPR_DISABLE_UI=${COOPR_DISABLE_UI:-false}
 export COOPR_RUBY=${COOPR_RUBY:-ruby}
 export COOPR_USE_DUMMY_PROVISIONER=${COOPR_USE_DUMMY_PROVISIONER:-false}
 export COOPR_API_USER=${COOPR_API_USER:-admin}
+export COOPR_API_USER_PASS=${COOPR_API_USER_PASS:-realtime}
```
albertshau: lets use 'password' instead of 'realtime'

albertshau:
structure of the code seems fine. Couple small comments and waiting on tests.

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
amakogoncv, line 64: I'm not sure about all settings, because some of them are used directly in the auth service and for now this service has many not universal parts. I hope, after refactoring in the cdap repo we will have such ability to use custom property names.

coopr-server/src/main/java/co/cask/coopr/runtime/ServerMain.java
amakogoncv, line 144: I think, that configuration classes will be moved to the common repo with auth service. Yes, this class, by default, loads cdap configs, because they are hard coded there, as you could see. I don't like this, the same, as you, but, unfortunately, auth service is using this class now. Hope, after refactoring and moving auth client to the common repo, this could be changed too.

coopr-standalone/bin/coopr.sh
amakogoncv: ok, will done

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
amakogoncv, line 64: @sreevatsanraman , could you note this, please. Maybe I'm mistaken.

coopr-server/src/main/java/co/cask/coopr/runtime/ServerMain.java
amakogoncv, line 144: @sreevatsanraman , could you note this too, please.

amakogoncv:
Locally all UTs are passed, but I got _Connection refused_ exception in the Travis CI. 
I think, that I specified incorrect testing port for the auth. service:

```
<property>
<name>security.auth.server.bind.port</name>
<value>5505</value>
</property>
```

@albertshau  If there is some available range of testing ports, please, let me to know.

amakogoncv:
Checked. Problem was with testing port.

coopr-server/README.md
```diff
@@ -0,0 +1,223 @@
+## Authentication support
```
albertshau: this is good, but I think it would be better to put it in the coopr-docs directory, as its own section in the installation guide.

amakogoncv: Ok, will done

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
albertshau, line 64: Ok I see, the SecurityAuthenticationHttpHandler takes in the CConfiguration object.  But since we're reading securityEnabled and realm in the coopr code and passing it into the SecurityAuthenticationHttpHandler, at least those 2 config settings can come from coopr specific setting names instead of cdap setting names.  @sreevatsanraman we should be sure to remove usage of the config when extracting that class into the common repo

coopr-server/src/main/java/co/cask/coopr/http/handler/auth/SecurityAuthenticationHttpHandler.java
```diff
@@ -0,0 +1,317 @@
...
+   *                                   external auth servers URIs
+   * @throws Exception
+   */
+  private void stopWatchWait(JsonArray externalAuthenticationURIs) throws Exception {
+    boolean done = false;
+    Stopwatch stopwatch = new Stopwatch();
+    stopwatch.start();
+    String protocol;
+    int port;
+    if (configuration.getBoolean(Constants.Security.SSL_ENABLED)) {
```
albertshau, line 171: @sreevatsanraman seems like these 3 settings should be final and passed into the constructor

coopr-server/src/test/java/co/cask/coopr/BaseTest.java
```diff
@@ -68,13 +73,16 @@
  * Base class with utilities for loading admin entities into a entityStore and starting zookeeper up.
  */
 public class BaseTest {
+  protected static final String TEST_COOPR_CONFIG = "coopr-test.xml";
+  protected static final String DEFAULT_COOP_CONFIG = "coopr-default.xml";
```
albertshau: COOP -> COOPR for consistency

coopr-server/src/test/java/co/cask/coopr/BaseTest.java
```diff
@@ -68,13 +73,16 @@
  * Base class with utilities for loading admin entities into a entityStore and starting zookeeper up.
  */
 public class BaseTest {
+  protected static final String TEST_COOPR_CONFIG = "coopr-test.xml";
```
albertshau: I don't think we should add conf files, but should just set the correct settings through the conf java apis.  Otherwise we set a pattern of adding conf files and can end up with a bunch of files all for different test scenarios.

albertshau: additionally, in the java we can use the constants for setting names which is better than just writing them out in conf files if we ever change the setting names.

coopr-server/src/test/java/co/cask/coopr/http/AuthenticationServerTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.net.URL;
+
+/**
+ * Test the External Authentication service integration
+ */
+public class AuthenticationServerTest extends ServiceTestBase {
+  private static final String AUTH_TEST_CONFIG = "coopr-test-auth.xml";
```
albertshau: similar comment about setting security settings directly in java instead of through files.

albertshau:
some comments, mostly about moving things around.  After that, lgtm.

coopr-server/src/test/java/co/cask/coopr/BaseTest.java
amakogoncv: ok, I will revert this back. I changed it, to avoid setting parameters to the both `Configuration` and `CConfiguration` instances. But, I still hope, that in the future `CConfiguration` will be removed and stay only one config, so it is not a problem

amakogoncv:
This test code could be simplified, after refactoring auth service, for use `Configuration` instead of `CConfiguration`.

coopr-server/src/main/java/co/cask/coopr/http/HandlerServer.java
amakogoncv, line 64: `CFG_REALM` - this one could be replaced with Coopr specific one, but, unfortunately, `CFG_SECURITY_ENABLED` - no, because it is also read by the cdap dependent code now.

albertshau, line 64: ok its fine then, lets just make a comment to change when we can.

amakogoncv:
@albertshau , Is it what you mean?

albertshau:
yep. this lgtm pending travis

amakogoncv:
@albertshau , Issue with unit tests in the Travis CI was resolved. Requests were sent earlier then authentication server starting process completed.

coopr-server/src/test/java/co/cask/coopr/http/AuthenticationServerTest.java
```diff
@@ -0,0 +1,163 @@
...
+    testServerPort = handlerServer.getBindAddress().getPort();
+    authURL = String.format("http://%s:%d/token",
+                            externalAuthenticationServer.getSocketAddress().getHostName(),
+                            externalAuthenticationServer.getSocketAddress().getPort());
+  }
+
+  private static void initAuthTestProps() throws IOException {
+    conf.setBoolean(Constants.Security.CFG_SECURITY_ENABLED, true);
+    conf.set(Constants.Security.AUTH_SERVER_ADDRESS, "127.0.0.1");
+    conf.setInt(Constants.Security.AUTH_SERVER_PORT, TestHelper.getFreePort());
```
albertshau, line 74: It looks like ExternalAuthenticationServer has a getSocketAddress method.  You should be able to set the port to 0, let it bind to a random port, then use that method to get the port it bound to. That way we eliminate the possible of a race where something binds to the "free" port between the time that you found it and the time the server tries to bind.

amakogoncv, line 74: I already checked it. When I specify port as 0, auth service really start in the any accessible port, but `getSocketAddress()` continue to retrieve 0 port. You could try to do this locally.

albertshau, line 74: Ok I see, that's unfortunate.  I will open a jira to expose the actual port, as this should be useful to whoever wants to use it in unit tests.

amakogoncv, line 74: Yes, you are right. What about this PR. Could it be merged, or you have some another comments?

coopr-server/src/test/java/co/cask/coopr/http/AuthenticationServerTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+  @BeforeClass
+  public static void setup() throws IOException, InterruptedException {
+    initAuthTestProps();
+    handlerServer = injector.getInstance(HandlerServer.class);
+    handlerServer.startAndWait();
+    externalAuthenticationServer = injector.getInstance(ExternalAuthenticationServer.class);
+    externalAuthenticationServer.startAndWait();
+    // Wait *SLEEP_TIME_IN_SEC* for the complete service start
+    Thread.sleep(1000 * SLEEP_TIME_IN_SEC);
```
albertshau, line 64: this is not great either, but I see the ExternalAuthenticationServerTest does the same thing... so I guess it's fine to have this here for now.

albertshau:
opened some jiras to address less than ideal behavior of dependent classes.  This lgtm.

PR #764, opened 2014-11-12
[ui_ssl_tls_improvements] coopr.sh and server.js imrovements

kpovetkin:
coopr.sh improved to show correct link in case of https startup. ui server.js improved to add tls data on every request.

### DIFF
--- a/coopr-ui/server.js
+++ b/coopr-ui/server.js
@@ -41,6 +41,9 @@ var BOX_ADDR = CLIENT_ADDR + '/v2';
 var CLIENT_DIR = env === 'production' ? 'client-built' : 'client';
 var REJECT_UNAUTH = argv.rejectUnauth === 'true';
 
+var TLS_ENABLED_SETTING = 'tlsEnabled';
+var AGENT_OPTIONS_SETTING = 'agentOptions';
+
 console.info('Environment:', env, BOX_ADDR, CLIENT_DIR);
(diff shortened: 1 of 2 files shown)


coopr-ui/server.js
```diff
@@ -187,11 +187,41 @@ site.app.use(function (err, req, res, next) {
...
+        try {
+            keyFile = fs.readFileSync(tlsKey);
+        } catch(e) {}
+
+        try {
+            certFile = fs.readFileSync(tlsCrt);
+        } catch(e) {}
+
+        try {
+            caFile = fs.readFileSync(tlsCA);
```
gaarf, line 219: :-1: this is bad, 3x `readFileSync` - is blocking

`updateRequestOptionsWithTls` is called in `getEntity` which is all over the place, so this would happen virtually for every request, possibly multiple times! I know this is the old UI but still...

TLS creds must be read from disk on app startup and then cached in memory.

wolf31o2:
Shell LGTM... we'll let UI guys approve the rest.

coopr-ui/server.js
yg2017, line 219: done.

yg2017:
@gaarf I have fixed your comment. Please review.

gaarf:
thanks @ygolub, this is better. LGTM

PR #765, opened 2014-11-12
Feature/update focus module

ajainarayanan:
- Updates references of `my-focus` to `cask-focus` in all features and directives
- Updates bower and main app to refer to the bower component instead of the local directive and service.


gaarf:
:+1:  LGTM pending travis :8ball:

ajainarayanan:
Had to run twice to have both Karma and Protractor pass. Will try this until we reach a point where it is impossible to have protractor. @nistala Trying this as experiment. If it is impending your work please let me know will revert back.

PR #767, opened 2014-11-13
Increase default timeout to 10 seconds

wolf31o2:
After discussion with @gaarf and the possibility of slow client networks and very large returned payloads, especially for the cluster listing page when logged in as admin, I am increasing the default timeout from 3 seconds to 10 seconds.


gaarf:
LGTM :+1:

wolf31o2:
Test failures are timeouts. Merging.

PR #772, opened 2014-11-13
COOPR-567: Split server APIs to separate ports.

ghost:


albertshau:
travis is showing compilation errors.  Please fix.

albertshau:
logic looks good to me.  But we should be able to configure both APIs differently.  For example, should be able to run the tasks API with 2-way tls but have no tls for the external api.

albertshau:
lgtm, but travis consistently fails on ngui section. @gaarf @ajainarayanan is this a known issue?

ghost:
PR updated according to last comments. Please review.

albertshau:
passed today... not sure what is going on with travis but it is very flaky.

PR #777, opened 2014-11-19
Feature/extend tests mod

ajainarayanan:
In addition to @nistala's additional test cases the following are the minor changes that have been introduced,
- Makes each test suite self sustained.
- Adds two helper methods for edit and delete of assets from a list (like list of templates, images, providers, services etc.,)
- Dynamically creates a new asset (provider or services or others) and edits and deletes them. Shouldn't depend on existing assets
- Removes small cases where we are checking for clean slate (count to be 0)
- Introduces a new reporter for slightly better logging of tests.

Currently (while this PR was about to be created https://travis-ci.org/caskdata/coopr/builds/41430085) the tests seems to pass.
We still need to ascertain that the changes introduced on top of @nistala's changes doesn't break existing stuff. (re-running the tests again and again to find anomalies)


coopr-e2e/test/e2e/hardwares/hardware-spec.js
```diff
@@ -36,22 +33,26 @@ describe('hardware types test', function () {
...
-    element(by.css('.modal-dialog .modal-footer .btn-primary')).click();
-    hardwareTypes = element.all(by.repeater('item in list'));
-    expect(hardwareTypes.count()).toEqual(0);
+    var selectedHardware;
+    hardwareNames = element.all(by.repeater('item in list').column("item.name"));
+    hardwareCount = hardwareNames.count();
+
+    helper.deleteAssetFromList(hardwareNames, hardware);
+
+    expect(hardwareCount.then(function(i) {
```
rns4731: can you do:
expect(hardwareNames.count).toEqual(hardwareCount -1)?

coopr-e2e/test/e2e/providers/provider-spec.js
```diff
@@ -55,19 +57,20 @@ describe('providers test', function () {
...
+    var providerNames;
 
     // List
-    providersList = element.all(by.repeater('item in list'));
-    expect(providersList.count()).toEqual(1);
-    var providerText = providersList.first().getText();
-    expect(providerText).toMatch('Testprovider Test description Edit\nDelete');
-    providersList.first().element(by.cssContainingText('.btn', 'Edit')).click();
+    providerNames = element.all(by.repeater('item in list').column('item.name'));
+    helper.editAssetFromList(providerNames, provider);
```
rns4731: should this check if there is 1 extra provider?

coopr-e2e/test/protractor-help.js
```diff
@@ -28,6 +32,44 @@ function login(tenant, username, password) {
 
 }
 
+function deleteAssetFromList(list, name) {
```
rns4731: looks like using these implicitly requires us to write things with "then" syntax for actions happening after since we don't know when these will resolve, would adding a waitForAngular or an expect in here resolve that?

coopr-e2e/test/e2e/services/service-spec.js
```diff
@@ -0,0 +1,95 @@
...
+    element(by.cssContainingText('.my-thing-picker a.btn-warning:not([disabled])', 'add feature')).click();
+    element(by.css('#newThingName-feature')).sendKeys('feature1');
+    element(by.partialButtonText('Add')).click();
+
+    // This assumes that there is a "base" service created by default
+    // and in addition to that we have added a service called "feature1" under provides.
+    var provides = element.all(by.repeater('name in model'));
+    expect(provides.count()).toEqual(2);
+
+    element(by.css('.btn.add-automator-action')).click();
```
rns4731, line 37: i had added the class because by.cssContainingText('.btn', 'add automator action') wasn't working, same for .action-button, is that still the case?

coopr-ngui/app/features/services/form.html
```diff
@@ -167,7 +167,7 @@
   <div class="row">
     <div class="col-xs-offset-2 col-xs-10">
       <p class="form-control-static">
-        <a href="" class="btn btn-info btn-xs"
+        <a href="" class="btn btn-info btn-xs add-automator-action"
```
gaarf, line 170: I thought I fixed this already in dev... avoid adding classes to markup purely for selenium targeting.

coopr-ngui/app/features/templates/form.html
```diff
@@ -39,7 +39,7 @@
               <span class="fa fa-long-arrow-right"></span>
           </button>
 
-          <button type="submit" class="btn btn-primary" 
+          <button type="submit" class="btn btn-primary action-btn"
```
gaarf, line 42: avoid adding classes to markup purely for selenium targeting.
using something like `element(by.css('form button > .fa-save')).click()`, should work.

coopr-e2e/test/protractor-conf.js
```diff
@@ -17,31 +17,38 @@ var config = {
...
 
   jasmineNodeOpts: {
-    defaultTimeoutInterval: 60000
+    defaultTimeoutInterval: 60000,
+    silent: true
   },
 
   onPrepare: function() {
     browser.driver.manage().window().maximize();
+    var SpecReporter = require('jasmine-spec-reporter');
```
rns4731, line 33: can we see total assertions?

coopr-e2e/test/protractor-help.js
```diff
@@ -28,6 +32,44 @@ function login(tenant, username, password) {
...
+  var selectedItem;
+  list
+    .then(function(s) {
+      s.forEach(function(item, index) {
+        item.getText().then(function(text) {
+          if (text === name) {
+            selectedItem = item;
+          }
+        });
+      });
```
gaarf: not critical, but why not  have the promise return (=== resolve with) something? that way can avoid the `selectedItem` var.

coopr-e2e/test/e2e/clusters/cluster-spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+  it('should log in', function () {
+    helper.loginAsAdmin();
+  });
+  
+  it('should go to cluster create page', function () {
+
+    browser.get('/clusters');
+    clusters = element.all(by.repeater('item in list'));
+    expect(clusters.count()).toEqual(0);
```
rns4731, line 20: remove 19 and 20?

rns4731:
reran tests a few times, looks like they're passing, except for comments above LGTM

coopr-e2e/test/protractor-help.js
```diff
@@ -28,6 +32,44 @@ function login(tenant, username, password) {
...
+    })
+    .then(function() {
+      selectedItem.element(by.xpath("ancestor::tr"))
+        .element(by.cssContainingText('.btn', 'Edit')).click();
+    });
```
gaarf: same as above... as I think about it, why not just:

```
list.element(by.cssContainingText('tr', name)).element(by.cssContainingText('.btn', 'Edit')).click();
```

?

coopr-e2e/test/e2e/hardwares/hardware-spec.js
ajainarayanan: So that will fail as hardwareCount is a promise before resolving to a value.

coopr-e2e/test/e2e/providers/provider-spec.js
ajainarayanan: We can add the check. I think I just replaced the edit logic. If it needs check if we have one extra provider then we can do that.

coopr-e2e/test/e2e/services/service-spec.js
ajainarayanan, line 37: I did not change any of the test case logic. Just made it a bit independent of existing states. We can improve the logic of the test cases over the period of time.

coopr-e2e/test/protractor-conf.js
ajainarayanan, line 33: I believe there is no option available in `jasmine-spec-reporter` to get total assertions but the summary of the test runs displays the total no.of specs ran (both success and failure). I thought each test case will have only one assertion and that would do only one thing. Currently I am not aware of any solution in this `jasmine-spec-reporter` reporter to display that.

coopr-ngui/app/features/templates/form.html
ajainarayanan, line 42: From earlier commits. Will fix them.

coopr-e2e/test/e2e/clusters/cluster-spec.js
ajainarayanan, line 20: I did not modify the cluster spec. Once we add more tests to it then I think we can prune the spec.

coopr-e2e/test/protractor-help.js
ajainarayanan: :+1:

coopr-e2e/test/e2e/clusters/cluster-spec.js
rns4731, line 20: i meant now since we're assuming we don't know the state of the system, we should remove these because you could have existing clusters locally

rns4731:
:+1:

coopr-e2e/test/protractor-conf.js
gaarf, line 33: please move the `require` call out of `onPrepare` and move it to the top of this file

PR #779, opened 2014-11-19
COOPR-600: CLI: doesn't show current user or tenant

ghost:
I have added current user_id and tenant_id to help information. If prompt require current user or tenant please let me know by reviewing this PR.


jwang47:
LGTM

PR #781, opened 2014-11-19
Coopr 607: Tweak usage api

ghost:


coopr-server/src/main/java/co/cask/coopr/metrics/MetricService.java
```diff
@@ -43,15 +40,16 @@
     year
   }
 
+  public enum TimeUnit {
```
albertshau: you can just use java's TimeUnit, which has a convert method for converting from one time unit to another. See https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/TimeUnit.html for details.

coopr-server/src/main/java/co/cask/coopr/metrics/MetricService.java
```diff
@@ -74,15 +72,17 @@ public TimeSeries getNodesUsage(ClusterTaskFilter filter) throws IOException {
...
-      return new TimeSeries(start != null ? start : 0, end != null ? end : System.currentTimeMillis(),
-                            Collections.<Interval>emptyList());
+      long startTime = start != null ? start : 0;
+      return validate(new TimeSeries(startTime, end != null ? end : System.currentTimeMillis(),
+                            Arrays.asList(new Interval(startTime))), timeunit);
     }
     List<ClusterTask> createTasks = getTasksByName(tasks, ProvisionerAction.CREATE);
     List<ClusterTask> deleteTasks = getTasksByName(tasks, ProvisionerAction.DELETE);
-    long startDate = start != null ? start : createTasks.isEmpty() ?
+    long startDate = start != null ? start * timeunit : createTasks.isEmpty() ?
```
albertshau: timeunit should just change the values returned, but have no effect on the start and end timestamps, or the timestamps in the intervals.  Timestamps should be in seconds since epoch.

coopr-server/src/main/java/co/cask/coopr/metrics/MetricService.java
```diff
@@ -139,6 +157,10 @@ private long getDeleteTaskTime(List<ClusterTask> deleteTasks, ClusterTask create
     return result;
   }
 
+  private long getAppropriateValue(double rawValue, long divider) {
```
albertshau: can use TimeUnit's convert method.

coopr-server/src/main/java/co/cask/coopr/metrics/MetricService.java
```diff
@@ -108,7 +108,25 @@ public TimeSeries getNodesUsage(ClusterTaskFilter filter) throws IOException {
...
+  }
+
+  /**
+   * Validates {@code rawTimeSeries} using specified {@code timeunit}
+   *
+   * @param rawTimeSeries the raw {@link TimeSeries}
+   * @param timeunit the {@link co.cask.coopr.metrics.MetricService.TimeUnit}
+   * @return validated {@link TimeSeries}
+   */
+  private TimeSeries validate(TimeSeries rawTimeSeries, long timeunit) {
```
albertshau: we shouldn't have to make another pass through the data in order to do this. Rather, should be using the correct time unit as we are creating the timeseries.

albertshau:
couple small comments, then lgtm

ghost:
PR updated according to last comments.

albertshau:
another couple comments then lgtm

ghost:
PR updated according to last comments.

ghost:
Updated co.cask.http:netty-http version from 0.6.0-SNAPSHOT to 0.7.0-SNAPSHOT.

albertshau:
lgtm pending travis

albertshau:
For future reference, please refrain from using snapshot in dependencies if at all possible.  For example, in this case it would be more appropriate to change to 0.6.0 instead of 0.7.0-SNAPSHOT.  Will merge this and change the dep separately.

PR #787, opened 2014-11-21
Coopr 567: Split server APIs to separate ports.

ghost:
I have removed ProvisionerHandler from External API and add this handler to Internal API. Also added 
"PUT /v2/plugins/providertypes/{providertype-id}"
"PUT /v2/plugins/automatortypes/{automatortype-id}"
requests to ProvisionerHandler and removed this requests from SuperadminHandler. For now Internal API handles TaskHandler and ProvisionerHandler. I have done some basic tests with coopr standalone - real cluster provisioning via UI.


coopr-server/src/main/java/co/cask/coopr/http/handler/ProvisionerHandler.java
```diff
@@ -229,6 +236,91 @@ public void getProviderTypeResource(HttpRequest request, HttpResponder responder
...
+          LOG.warn("Exception while closing request reader", e);
+        }
+      }
+    } catch (IllegalArgumentException e) {
+      responder.sendError(HttpResponseStatus.BAD_REQUEST, "invalid input: " + e.getMessage());
+    } catch (Exception e) {
+      responder.sendError(HttpResponseStatus.BAD_REQUEST, "invalid input");
+    }
+    return result;
+  }
```
jwang47, line 322: Looks like these methods are directly copied from `SuperadminHandler`, except the part that validates that the user is missing? Is that right? Is it done elsewhere now?

```
Account account = getAndAuthenticateAccount(request, responder);
-    if (account == null) {
-      return;
-    }
-    if (!account.isSuperadmin()) {
-      responder.sendError(HttpResponseStatus.FORBIDDEN, "user unauthorized, must be superadmin.");
-      return;
-    }
```

ghost, line 322: yes, it is. This methods used for handling provisioner requests. So they doesn't need user check.

jwang47, line 322: Isn't this external now, so we'll need user check?

ghost, line 322: This internal for now. Previously used by provisioner to communicate with Coopr server.

jwang47, line 322: Maybe I don't understand the code, but looks like this `ProvisionerHandler` is an external handler?

https://github.com/cybervisiontech/coopr/blob/coopr_567/coopr-server/src/main/java/co/cask/coopr/http/guice/HttpModule.java#L57

```
externalHandlerBinder.addBinding().to(ProvisionerHandler.class);
```

ghost, line 322: No, ProvisionerHandler is an internal handler. His methods handles Provisioner requests only.

jwang47, line 322: If so, then shouldn't `ProvisionerHandler` be added as as follows?

```
internalHandlerBinder.addBinding().to(ProvisionerHandler.class);
```

instead of this, which is currently in the PR

```
externalHandlerBinder.addBinding().to(ProvisionerHandler.class);
```

(or externalHandlerBinder is not named correctly)

ghost, line 322: yes, right

ghost:
PR updated according to last comments

coopr-server/src/main/java/co/cask/coopr/http/guice/HttpModule.java
```diff
@@ -39,24 +39,21 @@
...
+    externalHandlerBinder.addBinding().to(StatusHandler.class);
+    externalHandlerBinder.addBinding().to(RPCHandler.class);
     externalHandlerBinder.addBinding().to(SuperadminHandler.class);
-    externalHandlerBinder.addBinding().to(ProvisionerHandler.class);
     externalHandlerBinder.addBinding().to(PluginHandler.class);
+    externalHandlerBinder.addBinding().to(UserHandler.class);
+    externalHandlerBinder.addBinding().to(MetricHandler.class);
+
+    Multibinder<HttpHandler> internalHandlerBinder =
+      Multibinder.newSetBinder(binder(), HttpHandler.class, Names.named("Internal"));
```
jwang47: Could you make the `Names.named()` into constants? Makes it much easier to locate where it's used.

jwang47:
Just one more comment. Otherwise LGTM

jwang47:
LGTM. Merging when build passes.

PR #789, opened 2014-11-24
COOPR-568: changes for distributed mode

yg2017:


yg2017:
This update is according to the request https://github.com/caskdata/coopr-templates/pull/17#issuecomment-64243110

dereklwood:
CERT_PARAMETER will not be in the environment, so we need to determine the default the same as we do in `coopr-standalone/bin/coopr.sh`  (essentially read the config xml).  We will also need to update the COOPR_SERVER_URI to either http or https, also same as we do in `coopr-standalone/bin/coopr.sh`

yg2017:
@dereklwood I have updated setup.sh according to your comment. 
Please take a look.

coopr-provisioner/bin/setup.sh
```diff
@@ -2,15 +2,51 @@
...
+COOPR_SSL=`read_property server.ssl.enabled ${COOPR_SERVER_CONF}/coopr-site.xml`
+if [ "${COOPR_SSL}" == "true" ]; then
+  COOPR_PROTOCOL=https
+fi
+
+export COOPR_SERVER_URI=${COOPR_SERVER_URI:-${COOPR_PROTOCOL}://localhost:55054}
+
+export TRUST_CERT_PATH=`read_property server.ssl.trust.cert.path ${COOPR_SERVER_CONF}/coopr-security.xml`
+export TRUST_CERT_PASSWORD=`read_property server.ssl.trust.cert.password ${COOPR_SERVER_CONF}/coopr-security.xml`
+if [ -n TRUST_CERT_PATH ] && [ -n TRUST_CERT_PASSWORD ]; then
```
wolf31o2: These are not variables. Please, fix.

coopr-provisioner/bin/setup.sh
```diff
@@ -2,15 +2,51 @@
 
 # Configurable variables
 TIMEOUT=3
-export COOPR_SERVER_URI=${COOPR_SERVER_URI:-http://localhost:55054}
 export COOPR_API_USER=${COOPR_API_USER:-admin}
 export COOPR_API_KEY=${COOPR_API_KEY:-1234567890abcdef}
 export COOPR_TENANT=${COOPR_TENANT:-superadmin}
 export COOPR_HOME=${COOPR_HOME:-/opt/coopr}
+export COOPR_SERVER_CONF=${COOPR_HOME}/server/conf
```
wolf31o2: What if you're running `setup.sh` from a checkout, which is the use case I wrote this script to resolve?

coopr-provisioner/bin/setup.sh
```diff
@@ -2,15 +2,51 @@
...
+
+# set up TLS creds
+
+COOPR_PROTOCOL=http
+COOPR_SSL=`read_property server.ssl.enabled ${COOPR_SERVER_CONF}/coopr-site.xml`
+if [ "${COOPR_SSL}" == "true" ]; then
+  COOPR_PROTOCOL=https
+fi
+
+export COOPR_SERVER_URI=${COOPR_SERVER_URI:-${COOPR_PROTOCOL}://localhost:55054}
```
wolf31o2: Lines 14-37 are not necessary if COOPR_SERVER_URI has been set. There's no reason to require reading from the configuration file. The provisioner and the server can run on different machines. In this case, there is no `coopr-site.xml` file available.

coopr-provisioner/bin/setup.sh
```diff
@@ -2,15 +2,51 @@
...
+
+export TRUST_CERT_PATH=`read_property server.ssl.trust.cert.path ${COOPR_SERVER_CONF}/coopr-security.xml`
+export TRUST_CERT_PASSWORD=`read_property server.ssl.trust.cert.password ${COOPR_SERVER_CONF}/coopr-security.xml`
```
wolf31o2: A user should be able to specify TRUST_CERT_PATH and TRUST_CERT_PASSWORD rather than forcing the reading of `coopr-security.xml` as that assumes the provisioner and server are on the same host. In production, this is likely to not be the case. Our production Coopr installation has a set of Coopr Servers and a set of Coopr Provisioners. Provisioners can run on their own machines.

coopr-provisioner/bin/setup.sh
```diff
@@ -2,15 +2,51 @@
...
+if [ -n TRUST_CERT_PATH ] && [ -n TRUST_CERT_PASSWORD ]; then
+  export CERT_PARAMETER="--cert ${TRUST_CERT_PATH}:${TRUST_CERT_PASSWORD}"
```
wolf31o2, line 17: This will enable the `--cert` option even if a user specifies another COOPR_SERVER_URI.

coopr-provisioner/bin/setup.sh
wolf31o2: Also, this only reads the configuration under `server/conf` which is **not** the configuration read by the server in distributed mode.

wolf31o2:
How are you testing these changes? What files are you using in testing?

What if a user uses this script on a machine that has Coopr installed, but they're pointing to another machine using COOPR_SERVER_URI?

tyatskiv:
Working on it, will respond soon.

kpovetkin:
@wolf31o2 Updated according to your comments.

wolf31o2:
LGTM...

PR #792, opened 2014-11-24
Import cdap cookbook

wolf31o2:
This is a simple import-only PR... this _shouldn't_ cause any problems. The main areas of potential issue are changes to `build-essential` or the `yum` and `yum-epel` cookbooks.


duddun:
LGTM

PR #793, opened 2014-11-25
COOPR-621 Do not require Internet for chef-solo bootstrap

wolf31o2:
Here's the flow:

yum -> apt -> curl -> wget -> FAIL

This should allow organizations to provide the `chef` package via their own YUM/APT repositories, which may reside inside their firewall.


coopr-provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator.rb
```diff
@@ -147,30 +147,12 @@ def bootstrap(inputmap)
       generate_chef_primitive_tar(chef_primitive)
     end
 
-    log.debug "Attempting ssh into ip: #{ipaddress}, user: #{sshauth['user']}"
```
dereklwood, line 150: this line could be the only indication that coopr is doing anything in the case of ssh failure (depending on provider).  that was the original reason for it.  I think it could still be a useful debug message.

coopr-provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator.rb
```diff
@@ -294,6 +277,56 @@ def runchef(inputmap)
...
 
+  def install_chef(inputmap)
+    sshauth = inputmap['sshauth']
+    ipaddress = inputmap['ipaddress']
+
+    # do we need sudo bash?
+    sudo = 'sudo' unless sshauth['user'] == 'root'
+
+    write_ssh_file
+    @ssh_file = @task['config']['ssh-auth']['identityfile'] unless @ssh_keyfile.nil?
```
dereklwood, line 288: suggestion: @ssh_file is an instance var but used like a local var in each of the 3 methods.  Also looking at it, I don't think we need it at all, as the File.delete where it's used can just check @ssh_keyfile directly.

dereklwood:
one suggestion, otherwise :+1: lgtm

coopr-provisioner/worker/plugins/automators/chef_solo_automator/chef_solo_automator.rb
wolf31o2, line 288: I agree. However, I feel that should be a refactor across everywhere that it's used (fog_provider, shell_automator, too) rather than just a change here.

PR #794, opened 2014-11-25
[use_strict_ngui] Bundled application JS file wrapped for "use strict" usage

kpovetkin:


coopr-ngui/Gulpfile.js
```diff
@@ -132,6 +132,11 @@ gulp.task('js:app', function() {
        footer: '\n})('+PKG+');\n'
     }))
     .pipe(plug.concat('app.js'))
+    .pipe(plug.replace(/(^|\n)[ \t]*('use strict'|"use strict");?\s*/g, '$1'))
```
gaarf, line 135: if I read this correctly, it is just removing whitespace, which would be done by the minifier anyways.

coopr-ngui/Gulpfile.js
```diff
@@ -132,6 +132,11 @@ gulp.task('js:app', function() {
...
+    .pipe(plug.wrapper({
+       header: '(function(window, document, undefined) {\n\'use strict\';\n',
+       footer: '\n})(window, document);\n'
+    }))
```
gaarf, line 139: should we really be using `strict` for the entirety of app.js? Could there not be a case where we do _not_ want it enabled for a particular closure?

coopr-ngui/Gulpfile.js
kpovetkin, line 135: It removes 'use strict' or "use strict" from JS sources.

coopr-ngui/Gulpfile.js
kpovetkin, line 139: No, there could not be.
Strict mode makes browser to parse JS source code using ECMA standard and do not use custom(browser) features or features that are in test mode(ECMA v6 features).
More: 
1. http://www.ecma-international.org/ecma-262/5.1/ - See 'Strict mode restrictions' sections.
2. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode.

gaarf, line 139: @kpovetkin is very _strict_ about `use strict` :smile_cat:

coopr-ngui/Gulpfile.js
ajainarayanan, line 135: If I have some use case where I my string that is going to displayed in DOM has 'use strict' would this remove that too? I feel this would be an overkill. If developers introduce new file (or any js functionality) then we see if they have a `use strict` if not we enforce it in the review time. I feel this is a little intrusive. 

We wrap use strict in the IIFE that has any angular code and that should be fine. I guess.

coopr-ngui/Gulpfile.js
kpovetkin, line 139: yes, it's

coopr-ngui/Gulpfile.js
kpovetkin, line 135: It should do nothing with 'use string' occurrences in dom as long as html templates are not compiled into js modules. Even in case of templates compilation into js modules it's possible to design build process in such a way that any html templates modification would be avoided.

gaarf:
let's not do the regex search/replace, it will slow down the build and I don't think there's any instance of `use strict` that would benefit from removal anyways.

else LGTM :+1:

kpovetkin:
It would not too slow. '/(^|\n)[ \t]_('use strict'|"use strict");?\s_/g' is pretty simple, and do not require to much iterations. Regexp implementation in JS(moreover NodeJS implementation) is well optimized.

coopr-ngui/Gulpfile.js
ajainarayanan, line 135: @kpovetkin Yes we have usecase where we compile html templates as js modules (angular-template-cache). The string does not necessarily have to be in DOM alone, it could be in my directive as constant and might be displayed based on some condition. I am not saying we will have a use case of 'use strict' being a string I am just saying adding and removing that should be left to individual developers who are writing any module and should not be in the build process.

kpovetkin, line 135: 1. It's possible to add template compilation after this action. In that case templates would not be touched by the replacer.
2. It would not touch your constants. http://regex101.com/r/nP7wQ0/1
3. _Important_ 'use strict' is mandatory for a good JS developer: http://cjihrig.com/blog/javascripts-strict-mode-and-why-you-should-use-it/

gaarf:
I didn't say it would be _too_ slow, I said it would just be _slower_ and _unnecessary_.
We don't want the regex search/replace. Please remove it and let's merge this.

kpovetkin:
If replacing is not necessary this PR has no sense. The right way would be to wrap code that requires striict mode by hand. It could be closed.

PR #797, opened 2014-11-26
Feature/docker base

dereklwood:
- [x] import of docker cookbook 0.36.0.  0.36.0 doesn't have a metadata.rb so imported the previous version first for dependencies.
- [x] update coopr-templates submodule, as latest google centos6 images are required.


wolf31o2:
:+1: LGTM... Travis failure is UI-related...

PR #800, opened 2014-11-27
Remove coopr-ui

wolf31o2:
Now that `coopr-ngui` is our default, we're removing `coopr-ui` from the repository. Since the diff is too large to examine, I suggest looking at the individual commits. b68bc71 is the deletion of `coopr-ui` directory, only. bb1baef is the only code modifications, to be reviewed.

I will submit another PR after this to do the appropriate renaming of `coopr-ngui` to `coopr-ui` and subsequent code changes, for ease of review.


gaarf:
:hear_no_evil: kill it :+1:

PR #802, opened 2014-11-30
provide a way to leave iptables alone

dereklwood:
Check an attribute before doing anything with iptables.  Useful for cases where you want to leave iptables alone, not disable it.  No changes to default behavior.

### DIFF
--- a/coopr-provisioner/worker/plugins/automators/chef_solo_automator/resources/cookbooks/coopr_firewall/recipes/default.rb
+++ b/coopr-provisioner/worker/plugins/automators/chef_solo_automator/resources/cookbooks/coopr_firewall/recipes/default.rb
@@ -18,7 +18,11 @@
 #
 
 if node.key?('coopr_firewall')
-  include_recipe 'coopr_firewall::iptables'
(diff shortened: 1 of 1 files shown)


coopr-provisioner/worker/plugins/automators/chef_solo_automator/resources/cookbooks/coopr_firewall/recipes/default.rb
```diff
@@ -18,7 +18,11 @@
 #
 
 if node.key?('coopr_firewall')
-  include_recipe 'coopr_firewall::iptables'
+  if node['coopr_firewall'].key?('unmanaged') && node['coopr_firewall']['unmanaged'].to_s == 'true'
+    Chef::Log.info("Not managing host firewall per node['coopr_firewall']['unmanaged'] attribute")
```
wolf31o2, line 22: Should this be a log resource, instead?

wolf31o2:
One comment, otherwise :+1: LGTM

dereklwood:
discussed... ok for now as no compelling reason one way or another

PR #803, opened 2014-12-01
[coopr-573] auth token support added to UI.

kpovetkin:


gaarf:
this is the old coopr UI, which is being deprecated.

wolf31o2:
Actually, it's been removed. Boom!

PR #805, opened 2014-12-01
COOPR-624: Added dependency to slf4j-nop for correct bindings.

tyatskiv:
@wolf31o2 Dependency to slf4j-nop has been added for correct bindings, as mentioned in the https://issues.cask.co/browse/COOPR-624?focusedCommentId=12310&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12310

Please take a look.

### DIFF
--- a/coopr-cli/pom.xml
+++ b/coopr-cli/pom.xml
@@ -47,6 +47,11 @@
       <artifactId>coopr-rest-client</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


jwang47:
Please also add slf4j-api dependency here, so that the version doesn't get pulled from common-cli dependency.

tyatskiv:
@jwang47 The slf4j-api dependency has been added.

jwang47:
LGTM

PR #809, opened 2014-12-03
[coopr-573-ng] auth token implementation for ui

kpovetkin:


kpovetkin:
Up. Build failed due to build machine environment problems.

gaarf:
`cdap-auth-client` is not on npm yet, so this can not build ;-)

Thanks @kpovetkin for this reference implementation, it is very useful! But not something we can merge until we internally discuss the release and adoption of `cdap-auth-client` module.

PR #810, opened 2014-12-03
COOPR-623: Add usage of common-cli inbuilt commands

ghost:


jwang47:
LGTM. Merging after build passes.

PR #811, opened 2014-12-03
Use IP not localhost

wolf31o2:
The `cors-anywhere` module does some checking against the proxy hostname and it does not consider `localhost` to be valid. This is dumb. To work around this, we replace `localhost` with `127.0.0.1` which matches the `cors-anywhere` validity check for IPv4 addresses.

### DIFF
--- a/coopr-ui/server.js
+++ b/coopr-ui/server.js
@@ -23,9 +23,11 @@ var COOPR_HOME = process.env.COOPR_HOME || (__dirname + '/../'),
     COOPR_UI_KEY_FILE = process.env.COOPR_UI_KEY_FILE || COOPR_HOME + 'cert/server.key',
     COOPR_UI_CERT_FILE = process.env.COOPR_UI_CERT_FILE || COOPR_HOME + 'cert/server.crt',
     COOPR_CORS_PORT = parseInt(process.env.COOPR_CORS_PORT || 8081, 10),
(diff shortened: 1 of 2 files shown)


dereklwood:
coopr.sh :+1:  no clue about server.js :)

coopr-ui/server.js
```diff
@@ -23,9 +23,11 @@ var COOPR_HOME = process.env.COOPR_HOME || (__dirname + '/../'),
     COOPR_UI_KEY_FILE = process.env.COOPR_UI_KEY_FILE || COOPR_HOME + 'cert/server.key',
     COOPR_UI_CERT_FILE = process.env.COOPR_UI_CERT_FILE || COOPR_HOME + 'cert/server.crt',
     COOPR_CORS_PORT = parseInt(process.env.COOPR_CORS_PORT || 8081, 10),
-    COOPR_SERVER_URI = process.env.COOPR_SERVER_URI || 'http://127.0.0.1:55054',
+    COOPR_SERVER_URI = process.env.COOPR_SERVER_URI || 'http://127.0.0.1:55054';
 
-    color = {
+COOPR_SERVER_URI = COOPR_SERVER_URI.replace('/localhost:', '\/127.0.0.1:')
```
dereklwood, line 28: i dont know .js, but should both the `/` be escaped or not consistently?

gaarf, line 28: it's a string, not a regexp. so it does not need to be escaped.

`COOPR_SERVER_URI.replace('/localhost:', '/127.0.0.1:');`

PR #812, opened 2014-12-04
Run register in foreground

wolf31o2:
Otherwise, the script will not exit with the correct exit code on a registration failure.


duddun:
LGTM

PR #815, opened 2014-12-04
Use JAVACMD rather than path-default java to check Java version

wolf31o2:
Copy caskdata/cdap#741 to use JAVACMD... thanks @sumitsu


duddun:
LGTM

PR #819, opened 2014-12-05
Ship profile.d script for packages

wolf31o2:
This will set the PATH for using `coopr-cli.sh` from anywhere on the system.


duddun:
LGTM

PR #818, opened 2014-12-04
Remove coopr-integration-testing

wolf31o2:
This was for the old UI...

PR #820, opened 2014-12-05
Restore coopr-integration-testing

wolf31o2:
Restore `coopr-integration-testing`, per @albertshau on #818 as it contains tests for CLI and other components, too... meh


duddun:
LGTM

PR #821, opened 2014-12-05
COOPR-635 Rename 'Pause Deploy' button

wolf31o2:
This should be less confusing. Fixes https://issues.cask.co/browse/COOPR-635


jwang47:
LGTM

PR #823, opened 2014-12-05
Import hadoop_wrapper version 1.1.12

wolf31o2:
This should resolve COOPR-620


wolf31o2:
Import-only PR, so self-approving and merging.

PR #824, opened 2014-12-05
Import cdap version 1.1.0

wolf31o2:
This is an import-only PR...


wolf31o2:
Self-approving and merging, since it's import-only...

PR #825, opened 2014-12-05
manual import of latest haproxy2 from repo, fixes missing name

dereklwood:
importing haproxy2 cookbook from https://github.com/demonccc/chef-repo/tree/master/cookbooks/haproxy2

fixes the missing cookbook name in the metadata.  Now that opscode is apparently installing chef 12 clients by default, this is throwing an error


duddun:
LGTM

PR #813, opened 2014-12-04
[ui-build-to-maven] UI build process moved to maven

kpovetkin:

PR #827, opened 2014-12-05
Update build for UI so we do not build twice, per @gaarf

wolf31o2:


wolf31o2:
Closing this, as a different solution was merged via #813

PR #832, opened 2014-12-08
Import users version 1.7.0

wolf31o2:


wolf31o2:
This is import-only. Merging (and ignoring Travis errors)...

PR #834, opened 2014-12-08
Import chef-solo-search version 0.5.1

wolf31o2:


wolf31o2:
This is import-only. Merging (and ignoring Travis errors)...

PR #835, opened 2014-12-08
Support for adding users

wolf31o2:
This gives Coopr users the ability to provide a `users` data bag to the `users` cookbook. Any user in the `sysadmin` group will get added to the system. This is the same group that the `sudo` cookbook uses, so we are including that, too. We need `chef-solo-search` for the `users` cookbook to find the data bag items.


duddun:
LGTM (other than travis failure)

PR #836, opened 2014-12-09
Import hadoop version 1.8.1

wolf31o2:


wolf31o2:
This is import-only, so merging... :smile:

PR #837, opened 2014-12-09
Allow a way to disable the base cookbooks

wolf31o2:
Disable the built-in cookbooks dns, firewall, and hosts by settings `base['no_dns']`, `base['no_firewall']`, and `base['no_hosts']`, respectively.

**WARNING** Disabling these services may cause your cluster to fail


duddun:
LGTM

PR #839, opened 2014-12-09
COOPR-617: Resource versioning

ghost:
This PR also include changes for https://issues.cask.co/browse/COOPR-195.


coopr-server/src/main/java/co/cask/coopr/codec/json/current/AbstractBaseVersionedEntityCodec.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package co.cask.coopr.codec.json.current;
+
+import co.cask.coopr.spec.BaseEntity;
+import co.cask.coopr.spec.BaseVersionedEntity;
+import com.google.gson.*;
```
jwang47: Nit: Could you add each individual import? We typically don't use wildcards in imports.

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
```diff
@@ -193,6 +256,37 @@ public void getImageType(HttpRequest request, HttpResponder responder,
...
+  public void getImageType(HttpRequest request, HttpResponder responder,
+                           @PathParam("imagetype-id") String imagetypeId, @PathParam("version") String versionStr) {
+    Account account = getAndAuthenticateAccount(request, responder);
+    if (account == null) {
+      return;
+    }
+
+    int version = getVersionFromString(responder, versionStr);
+    if (version <= 0) {
+      return;
```
jwang47, line 280: Should respond with `400 Bad Request`.

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
```diff
@@ -193,6 +256,37 @@ public void getImageType(HttpRequest request, HttpResponder responder,
...
+   * @param request The request for the image type.
+   * @param responder Responder for sending the response.
+   * @param imagetypeId Id of the image type to get.
+   * @param versionStr Version of the image type to get.
+   */
+  @GET
+  @Path("/imagetypes/{imagetype-id}/{version}")
+  public void getImageType(HttpRequest request, HttpResponder responder,
+                           @PathParam("imagetype-id") String imagetypeId, @PathParam("version") String versionStr) {
+    Account account = getAndAuthenticateAccount(request, responder);
```
jwang47, line 273: It's unclear that `getAndAuthenticateAccount` sends a response iff `account == null`. Let's think of ways to refactor this to be more clear, but this can be done in a later PR.

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
jwang47, line 280: Just saw that `getVersionFromString` also sends a response in certain conditions. Let's think of a way to make this more clear.

For example, if we change the condition here (`version <= 0`), then we would also have to be careful and make sure to change the condition in `getVersionFromString`, which isn't good.

jwang47, line 280: However, this unclearness is sort of a symptom of `netty-http`'s API.

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
```diff
@@ -193,6 +256,37 @@ public void getImageType(HttpRequest request, HttpResponder responder,
...
+      return;
+    }
+
+    int version = getVersionFromString(responder, versionStr);
+    if (version <= 0) {
+      return;
+    }
+
+    try {
+      respondToGetEntity(entityStoreService.getView(account).getImageType(imagetypeId, version),
```
jwang47, line 284: Could `entityStoreService.getView(account)` be null? If so, let's do a check and send a proper response. Otherwise the client gets a vague error response from the handler.

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
```diff
@@ -1162,6 +1501,20 @@ public void importConfig(HttpRequest request, HttpResponder responder) {
...
+    }
+    if (version <= 0) {
+      responder.sendString(HttpResponseStatus.BAD_REQUEST, "Version must be higher then zero");
+    }
+    return version;
+  }
+
   private <T> T getEntityFromRequest(HttpRequest request, HttpResponder responder, Type tClass) {
     T result = null;
     try {
```
jwang47, line 1848: Please also verify in a unit test that the endpoints without version give the latest version of the resource.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/AbstractBaseVersionedEntityCodec.java
ghost: yes

coopr-server/src/main/java/co/cask/coopr/spec/BaseVersionedEntity.java
```diff
@@ -0,0 +1,76 @@
...
+
+package co.cask.coopr.spec;
+
+import co.cask.coopr.common.conf.Constants;
+import com.google.common.base.Objects;
+
+/**
+ * A base for entities that require a name, version and optionally support an icon, description, and label.
+ */
+public class BaseVersionedEntity extends BaseEntity {
```
jwang47: Are we still using `BaseEntity` after these changes?

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
ghost, line 273: ok

coopr-server/src/main/java/co/cask/coopr/spec/BaseVersionedEntity.java
jwang47: Let's simply add `version` field to `BaseEntity`.

ghost: Automatortypes and providertypes do not support versioning

coopr-server/src/main/java/co/cask/coopr/common/conf/Constants.java
```diff
@@ -180,4 +180,7 @@
   public static final String TENANT_HEADER = "Coopr-TenantID";
   public static final String SUPERADMIN_TENANT = "superadmin";
   public static final String ADMIN_USER = "admin";
+
+  public static final int FIND_MAX_VERSION = 0;
```
jwang47: Let's use -1 here to make it easier to understand, and not use 0.

ghost: ok

coopr-server/src/main/java/co/cask/coopr/store/entity/BaseEntityStoreView.java
```diff
@@ -97,22 +98,28 @@ public AutomatorType apply(@Nullable byte[] input) {
...
-    CLUSTER_TEMPLATE("clusterTemplate"),
-    PROVIDER_TYPE("providerType"),
-    AUTOMATOR_TYPE("automatorType");
+    PROVIDER("provider", true),
+    HARDWARE_TYPE("hardwareType", true),
+    IMAGE_TYPE("imageType", true),
+    SERVICE("service", true),
+    CLUSTER_TEMPLATE("clusterTemplate", true),
+    PROVIDER_TYPE("providerType", false),
+    AUTOMATOR_TYPE("automatorType", false);
```
jwang47: Let's version provider type and automator type too. In that case, everything is versioned.

ghost: ok

coopr-server/src/main/java/co/cask/coopr/store/entity/SQLAdminEntityStoreView.java
```diff
@@ -71,13 +83,34 @@ protected void deleteEntity(EntityType entityType, String entityName) throws IOE
         conn.close();
       }
     } catch (SQLException e) {
-      throw new IOException("Exception deleting entity of type " + entityType.name().toLowerCase()
+      throw new IOException("Exception deleting entities of type " + entityType.name().toLowerCase()
```
jwang47: Original version was good I think. You can only delete one entity here right?

coopr-server/sql/create-tables-mysql.sql
```diff
@@ -84,37 +84,42 @@ CREATE TABLE IF NOT EXISTS automatorTypes (
...
+    version BIGINT,
     tenant_id VARCHAR(64),
     service MEDIUMBLOB,
-    PRIMARY KEY (tenant_id, name)
+    PRIMARY KEY (tenant_id, name, version)
 ) ENGINE = InnoDB;
 
 CREATE TABLE IF NOT EXISTS clusterTemplates (
     name VARCHAR(255),
+    version BIGINT,
```
jwang47, line 121: Should we have a default value for version? Or is it handled by the server?

jwang47:
Please also include an upgrade tool that upgrades 0.9.8 database to 0.9.9 (tool should add a column to the schemas of the resource tables and set the version of existing resources to 1 or whatever the starting version is)

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
ghost, line 284: entityStoreService.getView(account) couldn't be null

coopr-server/src/main/java/co/cask/coopr/http/handler/AdminHandler.java
ghost, line 1848: done

coopr-server/src/main/java/co/cask/coopr/store/entity/SQLAdminEntityStoreView.java
ghost: No, here you delete all versions of some entity. Maybe, you want to delete latest version of this entity?

coopr-server/sql/create-tables-mysql.sql
ghost, line 121: It is handled by the server

ghost:
Addressed PR comments and added db upgrades tool.

coopr-server/src/main/java/co/cask/coopr/store/entity/SQLAdminEntityStoreView.java
jwang47: Then maybe "Exception deleting all versions of type `entity-type` with name `entity-name`"

coopr-server/sql/update-tables-mysql.sql
```diff
@@ -0,0 +1,33 @@
...
+ALTER TABLE services ADD version BIGINT AFTER name;
+ALTER TABLE clusterTemplates ADD version BIGINT AFTER name;
+
+UPDATE providerTypes SET version = 1;
+UPDATE automatorTypes SET version = 1;
+UPDATE providers SET version = 1;
+UPDATE hardwareTypes SET version = 1;
+UPDATE imageTypes SET version = 1;
+UPDATE services SET version = 1;
+UPDATE clusterTemplates SET version = 1;
```
jwang47: Put the ALTER TABLE and UPDATE statements in an if-block, so that only if providerType is missing version column do we add the version column and set all versions in the providerType rows to 1, and only if automatorType is missing version column .. and so on.

coopr-server/sql/update-tables-mysql.sql
```diff
@@ -0,0 +1,33 @@
+#
```
jwang47: Name this something like upgrade-tables-pre0.9.9-to-0.9.9. We may also want to have a bash script that detects the current version of the Coopr cluster and runs any upgrade scripts needed. However, we can do that in a later PR.

coopr-server/src/main/java/co/cask/coopr/store/entity/SQLAdminEntityStoreView.java
ghost: sounds good

coopr-server/sql/update-tables-mysql.sql
ghost: ok

ghost:
Updated PR

jwang47:
Let's wait for the partial templates to be merged so that we can add versioning to that. Changes so far LGTM.

ghost:
Added partial templates versioning

jwang47:
Please also add version field to `PartialTemplate`

ghost:
@jwang47 PartialTemplate extends AbstractTemplate; AbstractTemplate extends BaseEntity; BaseEntity already contains version field.

jwang47:
Oops. LGTM then.

jwang47:
Tests passed. Merging.

PR #840, opened 2014-12-09
COOPR-651 Update netty-http to 0.8.0

wolf31o2:
This is a prerequisite to COOPR-652


chtyim:
The change LGTM, but Travis failed, is it related?

duddun:
LGTM
Travail failure was due to an intermittent network blip when attempting to download packages or artifacts.

PR #841, opened 2014-12-10
[ui-build-maven-improvement] node_modules and bower_components removal commands added

kpovetkin:
Sometimes build process leads to corrupted UI build(Strict dependency injection problems could be seen browser console).


wolf31o2:
I would much prefer there be separate build and clean targets for the build. We should only clean out these directories on a `mvn clean` not on every build.

kpovetkin:
The reason why directories cleaning is in bash script is that on the second attempt to build an app artifact could be corrupted. It happens occasionally.

wolf31o2:
Correct. Which is why there needs to be a way to clean the directories, but it should be done via a `mvn clean` command. It's fine if this command calls a script. We should never just blow away directories.

wolf31o2:
Also, the clean should remove the generated `dist` directory, too.

kpovetkin:
Updated according to comments.

wolf31o2:
Build error is related to partials in server, so I am going to go ahead and merge this...

PR #842, opened 2014-12-10
[coopr-336] description field added to "cluster create" page

kpovetkin:
According to https://issues.cask.co/browse/COOPR-336


gaarf:
LGTM :+1:

PR #843, opened 2014-12-10
Update Coopr change log for unreleased (0.9.9) merged pull requests

duddun:


wolf31o2:
Let's go through this list today and clear out a bunch of things that don't need to make it to the CHANGELOG...

wolf31o2:
Let's just merge this so we don't lose track of them. We'll need to update it with the latest merges, anyway.

PR #844, opened 2014-12-11
[coopr-647] template name instead of template description is used on cluster creation page

kpovetkin:


gaarf:
Code does more than address https://issues.cask.co/browse/COOPR-647 ... IMHO the issue needs more discussion before code change.

kpovetkin:
More? Could you clarify, please?

gaarf:
@kpovetkin your edits to `form-ctrl.js` change the assumptions about "base" template. Your suggestion is probably good but it's not directly related to the issue being addressed.

The edit to `clusters/form.html` is not a substitute for the conversation that is now taking place in JIRA. We need to discuss UX changes (on JIRA...) before rushing to implement them.

PR #845, opened 2014-12-11
[coopr-648] dropdown control hides in "Service Set" block in case if there are no available actions.

kpovetkin:
According to https://issues.cask.co/browse/COOPR-648


gaarf:
code LGTM :+1:

wolf31o2:
@gaarf is this good to merge, then?

gaarf:
tests didn't pass... Little green checkmarks soothe my anxiety.

wolf31o2:
Failure was due to provisioner due to a new rubocop. I'll fix that and send a PR for it, but this should be safe to merge (all server/UI tests passed)...

wolf31o2:
Rubocop fixes are in #852

PR #847, opened 2014-12-12
COOPR:629 Update template APIs to support extending templates and including partials

rpedanov:
@jwang47 Branch ready for review.


rpedanov:
@jwang47 I'm not sure why build fails, it looks like some npm problems. Could you check it please?

coopr-server/src/main/java/co/cask/coopr/spec/template/Administration.java
```diff
@@ -20,10 +20,10 @@
 /**
  * Defines settings for cluster administration like lease duration, etc.
  */
-public class Administration {
+public final class Administration {
   public static final Administration EMPTY_ADMINISTRATION = new Administration(LeaseDuration.FOREVER_LEASE_DURATION);
 
-  private final LeaseDuration leaseDuration;
+  LeaseDuration leaseDuration;
```
jwang47: Why remove `private final`?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
```diff
@@ -16,32 +16,28 @@
...
+  private String dnsSuffix;
+  private JsonObject config;
 
   private ClusterDefaults(Set<String> services, String provider, String hardwaretype,
                           String imagetype, String dnsSuffix, JsonObject config) {
-    Preconditions.checkArgument(services != null && !services.isEmpty(), "default services must be specified");
-    //TODO: move it to the final phase
-//    Preconditions.checkArgument(provider != null, "default provider must be specified");
-    Preconditions.checkArgument(dnsSuffix == null || StringUtils.isValidDNSSuffix(dnsSuffix),
-                                dnsSuffix + " is an invalid DNS suffix.");
```
jwang47, line 44: Why remove these checks?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
```diff
@@ -129,12 +157,12 @@ public static Builder builder() {
     private JsonObject config = new JsonObject();
 
     public Builder setServices(Set<String> services) {
-      this.services = ImmutableSet.copyOf(services);
+      this.services = services == null ? Sets.<String>newHashSet() : services;
```
jwang47: `services == null ? ImmutableSet.of() : ImmutableSet.copyOf()`

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
```diff
@@ -129,12 +157,12 @@ public static Builder builder() {
...
       return this;
     }
 
     public Builder setServices(String... services) {
-      this.services = ImmutableSet.copyOf(services);
+      this.services = new HashSet<String>(Arrays.asList(services));
```
jwang47: Why change this to be mutable set?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -97,6 +110,122 @@ public ClusterTemplate build() {
...
+      private final Set<String> immutables = Sets.newHashSet();
+
+      public Merger setInitialTemplate(ClusterTemplate initialTemplate) {
+        parent = initialTemplate.parent;
+        includes = initialTemplate.includes;
+
+        return this;
+      }
+
+      public Merger merge(Collection<AbstractTemplate> templates) throws TemplateImmutabilityException {
```
jwang47: Should be `List<AbstractTemplate>` since order is important.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -97,6 +110,122 @@ public ClusterTemplate build() {
...
+        for (AbstractTemplate template : templates) {
+          if (template instanceof ClusterTemplate) {
+            copyFullProps(template);
+          } else {
```
jwang47: Should be `else if (template Instanceof PartialTemplate)`, and throw in the `else` - if we have a new child of `AbstractTemplate`, then we won't get unexpected behavior.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -97,6 +110,122 @@ public ClusterTemplate build() {
...
+        return merge(Arrays.asList(template));
+      }
+
+      private void validateFields() {
+        Preconditions.checkArgument(clusterDefaults.getServices() != null && !clusterDefaults.getServices().isEmpty(),
+                                    "default services must be specified");
+        Preconditions.checkArgument(clusterDefaults.getProvider() != null, "default provider must be specified");
+        Preconditions.checkArgument(clusterDefaults.getDnsSuffix() == null ||
+                                      StringUtils.isValidDNSSuffix(clusterDefaults.getDnsSuffix()),
+                                    clusterDefaults.getDnsSuffix() + " is an invalid DNS suffix.");
```
jwang47: Ok, so the checks are moved here.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -73,10 +83,13 @@ public Parent getParent() {
...
    */
   public static class Builder extends AbstractTemplate.Builder<ClusterTemplate, Builder> {
-
     private Parent parent;
     private Set<Include> includes = ImmutableSet.of();
 
+    public Merger merger() {
+      return new Merger();
+    }
+
```
jwang47: I don't think `Merger` should come from a builder method. What if we want to use multiple `Merger` implementations? I think we should instead have `Merger` be injected into whatever classes want to merge templates.

coopr-server/src/main/java/co/cask/coopr/spec/template/Compatibilities.java
```diff
@@ -133,37 +135,37 @@ public static Builder builder() {
    * Builder for creating Compatibilities.
    */
   public static class Builder {
-    private Set<String> hardwaretypes = ImmutableSet.of();
-    private Set<String> imagetypes = ImmutableSet.of();
-    private Set<String> services = ImmutableSet.of();
+    private Set<String> hardwaretypes = Sets.newHashSet();
+    private Set<String> imagetypes = Sets.newHashSet();
+    private Set<String> services = Sets.newHashSet();
```
jwang47: Why change these fields to be mutable sets?

coopr-server/src/main/java/co/cask/coopr/spec/template/Constraints.java
```diff
@@ -28,12 +28,12 @@
...
   private final Map<String, ServiceConstraint> serviceConstraints;
   private final LayoutConstraint layoutConstraint;
-  private final SizeConstraint sizeConstraint;
+  SizeConstraint sizeConstraint;
 
   public Constraints(Map<String, ServiceConstraint> serviceConstraints, LayoutConstraint layoutConstraint,
                      SizeConstraint sizeConstraint) {
     this.serviceConstraints = serviceConstraints == null ?
-      ImmutableMap.<String, ServiceConstraint>of() : serviceConstraints;
+      Maps.<String, ServiceConstraint>newHashMap() : serviceConstraints;
```
jwang47: Why change this?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -97,6 +110,122 @@ public ClusterTemplate build() {
...
+        //merge base entity properties
+        if (isNotBlank(from.getName())) name = from.getName();
+        if (isNotBlank(from.getLabel())) label = from.getLabel();
+        if (isNotBlank(from.getIcon())) icon = from.getIcon();
+        if (isNotBlank(from.getDescription())) description = from.getDescription();
+      }
+
+      @SuppressWarnings("unchecked")
+      private void mergeSet(Set dest, Set from) {
+        if (dest != null && from != null) {
```
jwang47: We made a lot of things mutable sets due to how we're merging the template here, right?

jwang47: Since we made a lot of things mutable sets/maps, let's move the immutable copy to the `build()` method of the builder, so that a built object doesn't use mutable sets/maps.

coopr-server/src/test/java/co/cask/coopr/spec/template/PartialTemplateTest.java
```diff
@@ -2,23 +2,30 @@
...
+import org.junit.runners.MethodSorters;
 
 import java.io.InputStream;
+import java.util.Map;
+import java.util.Set;
 
 /**
  *
  */
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
jwang47, line 29: Why is this needed?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
```diff
@@ -97,6 +110,122 @@ public ClusterTemplate build() {
...
+        mergeSet(compatibilities.getHardwaretypes(), from.getCompatibilities().getHardwaretypes());
+        mergeSet(compatibilities.getImagetypes(), from.getCompatibilities().getImagetypes());
+        mergeSet(compatibilities.getServices(), from.getCompatibilities().getServices());
+
+        //merge admin lease duration
+        if (!from.getAdministration().getLeaseDuration().equals(LeaseDuration.FOREVER_LEASE_DURATION)) {
+          administration.leaseDuration = from.getAdministration().getLeaseDuration();
+        }
+
+        //merge base entity properties
```
jwang47: We don't want to merge these. Instead, we'll use whatever is immediately provided. E.g. with user JSON, we would simply use what is provided in that JSON; if it's from a specified template, then it's from that template.

jwang47:
Please resolve merge conflicts

coopr-server/src/main/java/co/cask/coopr/spec/template/Administration.java
rpedanov: For merging purposes. We have to replace leaseDuration at the final merging process.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
rpedanov, line 44: Since we have partial templates this fields can be empty. So I moved this checks into Merger.validateFields() that it can be validated when the template will be merged.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
rpedanov: Since we have to merge several templates we have to merge all inner collections of this templates. For example if we have two templates with "defaults" for merge then we have to merge "services" set from second defaults into the first one. If set will be immutable then this operation will be no successfully.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterDefaults.java
rpedanov: See comment above

coopr-server/src/test/java/co/cask/coopr/spec/template/PartialTemplateTest.java
rpedanov, line 29: For test ordering. I want that templates will be parsed and persisted at the begin so we can get it in the next tests, and deleted at the final test method.

coopr-server/src/main/java/co/cask/coopr/spec/template/Constraints.java
rpedanov: For merging purposes. See comment above.

coopr-server/src/main/java/co/cask/coopr/spec/template/Compatibilities.java
rpedanov: For merging purposes. See comment above.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
rpedanov: Now this fields will be set with the values of last template that be provided in merge set. I mean that if we have something like that:
Child -> Parent -> Includes
Then values of this fields will be replaced with child values because we merge child at final. This allow to avoid some null value from the child. For example if parent has Label field value and child has not, then Label will be taken from parent. Or we should take this fields only from the last template(specified template) and set null values if values does not provided?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
rpedanov: LinkedHashSet always passed here. I believe that it saves insert order as well.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
jwang47: Let's take these fields only from the last template.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
jwang47: @rpedanov Please look at this comment.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
jwang47: @rpedanov Please look at this comment too.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
rpedanov: Hey @jwang47 , as I understand I have to remove all Mutable collections and its setters from builders? But create setters for mutable collections inside template entities (ClusterDefaults, Compatibilities)?  So if we try to create some entity we have to construct it from builder first and after that init not mutable fields through setter of this entity?

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
rpedanov: @jwang47 Yes, I already did it. I'm waiting when I done with moving mutable collections. Please, see my question below.

coopr-server/src/main/java/co/cask/coopr/spec/template/ClusterTemplate.java
jwang47: Right, let's keep the collections of the template entities (ClusterDefaults, Compatibilities) immutable as they were before. Just don't want to allow the possibility of any unexpected bugs.

rpedanov:
Immutable collections came back to template entities, merger separate service now. @jwang47 please check it.

coopr-server/src/main/java/co/cask/coopr/cluster/ClusterService.java
```diff
@@ -732,4 +741,85 @@ private Cluster getCluster(String clusterId, Account account) throws IOException
     }
     return cluster;
   }
+
+  /**
+   * Build cluster template from provided includes and parents.
+   *
+   * @param clusterTemplate Cluster template which can contains includes and parents.
+   * @param account Account of the user that is trying to resume a cluster job.
```
jwang47: Account of the user that is trying to resolve a template.

coopr-server/src/main/java/co/cask/coopr/cluster/ClusterService.java
```diff
@@ -732,4 +741,85 @@ private Cluster getCluster(String clusterId, Account account) throws IOException
...
+    throws TemplateNotFoundException, TemplateImmutabilityException, IOException {
+    EntityStoreView entityStore = entityStoreService.getView(account);
+    return resolveTemplate(entityStore, clusterTemplate);
+  }
+
+  /**
+   * Build cluster template from provided includes and parents.
+   *
+   * @param templateName Cluster template name.
+   * @param account Account of the user that is trying to resume a cluster job.
```
jwang47: Account of the user that is trying to resolve a template.

coopr-server/src/main/java/co/cask/coopr/cluster/ClusterService.java
```diff
@@ -732,4 +741,85 @@ private Cluster getCluster(String clusterId, Account account) throws IOException
...
     return cluster;
   }
+
+  /**
+   * Build cluster template from provided includes and parents.
+   *
+   * @param clusterTemplate Cluster template which can contains includes and parents.
+   * @param account Account of the user that is trying to resume a cluster job.
+   * @return Cluster Template with merged body from includes and parents.
+   * @throws IOException if there was some error writing to stores.
```
jwang47, line 751: We don't write to stores here, right? Just read?

coopr-server/src/main/java/co/cask/coopr/cluster/ClusterService.java
```diff
@@ -732,4 +741,85 @@ private Cluster getCluster(String clusterId, Account account) throws IOException
...
+   * @throws TemplateNotFoundException if template can't be found in store.
```
jwang47: The template is provided, so why do we need to throw TemplateNotFoundException?

jwang47:
Let's also add test cases for user providing an entire template rather than using a stored template. Can test the following:
- Provided template without error
- Provided template that tries to override immutable partial
- Provided template that tries to override immutable parent template
- Provided template that is badly formatted (i.e. not proper JSON)
- etc

(also make sure we have the same set of test cases that use stored templates)

Otherwise, the code changes mostly looks good.

rpedanov:
@jwang47 please take a look

jwang47:
Just one minor comment.

rpedanov:
@jwang47 new line added

PR #849, opened 2014-12-13
Import hadoop version 1.9.1

wolf31o2:


wolf31o2:
Since this is just an import, merging to develop...

PR #850, opened 2014-12-13
Import hadoop_wrapper version 0.1.13

wolf31o2:


wolf31o2:
Since this is an import-only PR, I am merging it.

PR #852, opened 2014-12-15
Feature/cleanups

wolf31o2:
- Rename 3 files which had "invalid" file names
- Fix a couple variable checks
- Add braces around variables


duddun:
LGTM

PR #856, opened 2014-12-17
Feature/minor script cleanups

wolf31o2:
Just a few cleanups of scripts and files. The only functionality change here is that we're now checking that the file listed in TRUST_CERT_PATH is now a readable file on the filesystem.


duddun:
LGTM

PR #857, opened 2014-12-18
update fog from 1.21.0 to 1.26.0

duddun:
This will ensure we have the fog 1.26.x gem installed, where x>=0.


dereklwood:
:+1: (but check travis)

wolf31o2:
Travis error appears to be transient.

duddun:
https://issues.cask.co/browse/COOPR-736

PR #865, opened 2014-12-20
Cleanup continuuity and invalid dash characters

wolf31o2:
There are several mentions of `continuuity` in the new partials code. Removing this. Also, there's a **ton** of places where the '-' (dash) character is some other encoding. Fixing these...


gaarf:
циісоↁэ іѕ fци! LБГМ :+1:

PR #866, opened 2014-12-22
Mix watch server start in gulp

kpovetkin:


coopr-ui/package.json
```diff
@@ -7,7 +7,6 @@
     "url": "http://github.com/caskdata/coopr.git"
   },
   "scripts": {
-    "start": "node ./server.js",
```
gaarf, line 10: don't remove `npm start`, it still works and could be needed elsewhere. It's better to be explicit than to rely on npm's default behavior.

coopr-ui/Gulpfile.js
```diff
@@ -293,3 +293,24 @@ gulp.task('watch', ['build'], function() {
...
+      './test/*',
+      './bower_components',
+      './node_modules',
+      'Gulpfile.js'
+    ],
+    env: {
+      'NODE_ENV': 'development'
+    },
+    nodeArgs: ['--debug']
+  })
```
gaarf: missing semicolon;

coopr-ui/Gulpfile.js
```diff
@@ -293,3 +293,24 @@ gulp.task('watch', ['build'], function() {
...
+});
+
+gulp.task('serve', ['watch', 'develop-server']);
```
gaarf, line 316: IMHO the name of the task should more obviously reflect it's intended use by FE developers. `gulp serve` does not imply that it will autobuild / autorestart on changes. I suggest `gulp develop`.

coopr-ui/Gulpfile.js
```diff
@@ -293,3 +293,24 @@ gulp.task('watch', ['build'], function() {
...
+    ignore: [
+      './app/*',
+      './dist/*',
+      './test/*',
+      './bower_components',
+      './node_modules',
+      'Gulpfile.js'
+    ],
+    env: {
+      'NODE_ENV': 'development'
```
gaarf: we are not using NODE_ENV anywhere in our code, what will be the effect of setting this here?

kpovetkin: There would not be any effect.

coopr-ui/Gulpfile.js
kpovetkin, line 316: It's done in this way because in Yeoman there is such task.

kpovetkin:
Updated

kpovetkin:
@gaarf Updated.

gaarf:
@ajainarayanan what do you think, do we want this? I have my muscle memory set to `gulp watch` and I like having the output in different tabs, so I'd never use it. I guess it doesnt hurt though?

coopr-ui/README.md
```diff
@@ -25,6 +25,7 @@ then, each in their own tab:
 
 * `gulp watch` (autobuild + livereload)
 * `npm start` (http-server + cors-anywhere)
+* `gulp serve` (autobuild + livereload + http-server)
```
gaarf: update the whole readme while you're at it... cors-anywhere is not used anymore. and if dev uses `gulp serve`, there no need for multiple tabs anymore.

kpovetkin:
@gaarf Updated

gaarf:
LGTM

PR #869, opened 2014-12-25
COOPR-234: Added version command to CLI.

tyatskiv:
@albertshau The version command has been added to the COOPR CLI in the same way, as it implemented for the CDAP.

### DIFF
--- a/coopr-cli/src/main/java/co/cask/coopr/shell/command/VersionCommand.java
+++ b/coopr-cli/src/main/java/co/cask/coopr/shell/command/VersionCommand.java
@@ -47,6 +47,6 @@ public class VersionCommand implements Command {
 
   @Override
   public String getDescription() {
-    return "Prints the version";
+    return "Prints the CLI version";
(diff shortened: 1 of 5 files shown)


coopr-cli/src/main/java/co/cask/coopr/shell/command/VersionCommand.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public String getPattern() {
+    return "version";
+  }
+
+  @Override
+  public String getDescription() {
+    return "Prints the version";
```
albertshau: It is more precise to describe this as the "CLI version", as it is possible to server and provisioner are running on a different version.

albertshau:
small comment, otherwise lgtm

tyatskiv:
@albertshau I have updated PR, please take a look.

albertshau:
:+1:

PR #870, opened 2014-12-26
COOPR-372: On plugin resource upload, check if the resource already exists

ghost:


ghost:
In this PR I have added functionality for checking if the resource already exists, and then return if it already exists instead of adding a new version.

wolf31o2:
This checks that it exists how? This needs to be more robust. The idea was that checking is done using a checksum, submitted by the client at upload time, stored in resource metadata, and checked by the server.

What is this supposed to do?

rpedanov:
@wolf31o2 So checksum must be provided as argument of rest service that client should send in request? Or checksum should be generated by the server when request received based on upload time? For example if same resource will be again uploaded in first 1 hour it will be not changed, otherwise version will be incremented. Is that correct scenario?

albertshau:
I think the idea was more that we would store a checksum as part of the resource metadata, and uploads that match the existing checksum would just be ignored instead of creating a new version of the resource.

wolf31o2:
Albert is correct.

The checksum should be provided by the client as data with the file, correct. However, it cannot be mandatory, due to backwards compatibility reasons. This was actually a part of a larger "resource verification" story, which is why we're missing some context. To facilitate this, several changes need to be made:
- data-uploader.rb should checksum (md5sum?) resources after they are assembled (tarball created, for "archive" resources) and send that checksum along with the resource to the server
- server should accept md5sum, if present... once the resource is uploaded, it should checksum the uploaded resource, if it does not match the provided checksum, the server should reject it due to file mismatch
- server should verify its calculated md5sum with previous versions of the resource, if it matches a previous version, we should ignore the upload and return the version of the matching resource
- server should store checksum metadata for any uploaded and stored versions

albertshau:
Do we want the client to provide the checksum or have the server calculate it? It makes more sense to me for the server to calculate it and keep it transparent to clients, though I haven't thought through everything.

wolf31o2:
Hrrrmn... you're correct. I was thinking that the client could optionally provide the checksum, which would allow the server to reject sooner, but I don't really see how that would work. Either the server calculates it and ignores anything from the client, or the server requires the client to provide this data and does verification on it. I guess just server-side would be the step towards full validation without changing client requirements.

jwang47:
Reopened PR here: https://github.com/caskdata/coopr/pull/891

PR #871, opened 2014-12-29
Coopr-300: Move lease times away from milliseconds

artemmakogon:
@jwang47 Please review


wolf31o2:
What about backwards compatibility with templates that use milliseconds? Is there a caskdata/coopr-templates PR to coincide with this?

artemmakogon:
Support of milliseconds has not been removed. It's possible to setup lease times in two ways: milliseconds and in string format

wolf31o2:
My bad. I missed where you're converting the strings to milliseconds internally. :+1:

jwang47:
`LeaseDuration` is within the JSON right? In that case, could you also add some tests that use gson to serialize and deserialize JSON into `LeaseDuration` objects with tests using milliseconds and tests using the new timestamp format?

PR #873, opened 2014-12-31
[coopr-539] info log on CREATE, DELETE events for provisioner added.

kpovetkin:
According to https://issues.cask.co/browse/COOPR-539

### DIFF
--- a/coopr-provisioner/worker/provisioner.rb
+++ b/coopr-provisioner/worker/provisioner.rb
@@ -124,14 +124,18 @@ end
 
 def _run_plugin(clazz, env, cwd, task)
   clusterId = task['clusterId']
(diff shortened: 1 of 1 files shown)


wolf31o2:
There's nothing wrong with these, but they should include a bit more information, possibly the hostname. Currently, you'll get the exact same message multiple times for a multinode cluster, which makes it a bit useless.

For create, something more like:

``` ruby
log.info "Creating node #{hostname} on #{provider} for #{cluster} using #{image} on #{hardware}"
```

For delete, just adding the host to the current messaging should suffice.

kpovetkin:
@wolf31o2 Updated.

wolf31o2:
:+1: LGTM... Thanks!

PR #874, opened 2014-12-31
Dockerfile for coopr-standalone

wolf31o2:


Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+EXPOSE 55054
+
+# Clean UP (reduce space usage of container as much as possible)
+RUN apt-get purge -y maven unzip build-essential git && \
+    apt-get autoclean && \
+    apt-get -y autoremove 
+
+ENV JAVA_HOME /usr/lib/jvm/java-6-openjdk-amd64
+
+# start CDAP in the background and tail in the foreground
```
duddun: We want to change CDAP to Coopr in the comment line.

Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+ENTRYPOINT /Software/coopr-[0-9]*.[0-9]*.[0-9]*-standalone/bin/coopr.sh start && \
```
duddun, line 81: This is the preferred form of ENTRYPOINT (exec):
ENTRYPOINT ["executable", "param1", "param2"](the preferred exec form)

In the other form (shell), ENTRYPOINT is run as a subcommand of 'sh -c', so executable does not have PID=1 and cannot not receive a sigterm (e.g. from 'docker stop ...').

Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+#
+# Cask is a trademark of Cask Data, Inc. All rights reserved.
+###############################################################################################
+# Please visit Docker.com and follow instructions to download Docker SW in your environment.
+# This Dockerfile will build a Coopr image from scratch utilizing ubuntu 12.04 as a base image.
+# The assumption is that you are running this from the root of the coopr directory structure.
+#
+FROM ubuntu:12.04
+MAINTAINER Cask Data <ops@cask.co>
+RUN apt-get update && \
```
duddun, line 23: I read some best practice info for Dockerfile, which suggests doing something like this: (keep each pkg on its own line)
RUN apt-get update && apt-get install -y \
   pkg 1 && \
   pkg 2 && \
   pkg 3 && \
   ...

Dockerfile
wolf31o2, line 81: You cannot do a compound command with the exec syntax.

Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+    apt-get install -y software-properties-common python-software-properties && \
+    add-apt-repository ppa:chris-lea/node.js && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends openjdk-6-jdk && \
+    apt-get install -y curl nodejs maven unzip ruby1.9.3 build-essential git
+
+RUN gem install fog --version 1.26.0 --no-rdoc --no-ri && gem install sinatra --version 1.4.5 --no-rdoc --no-ri && \
```
duddun: For readability, we should put each gem install command on its own line:
RUN gem install fog --version 1.26.0 --no-rdoc --no-ri && \
         gem install sinatra --version 1.4.5 --no-rdoc --no-ri && \
         gem install thin --version 1.6.2 --no-rdoc --no-ri && \
         gem install rest_client --version 1.7.3 --no-rdoc --no-ri && \
         gem install google-api-client --version 0.7.1 --no-rdoc --no-ri

Dockerfile
duddun, line 81: Ahh.

Dockerfile
wolf31o2, line 23: They also recommend running apt-get -y on each line. That's garbage. There's 0 reason to force us to do dependency resolution so many times. Many documented "best practices" are really very poor practices that waste time and energy, in the name of some individual's interpretation of "readability" improvements.

duddun:
It looks good so far.  Still looking, though.

duddun:
LGTM

PR #879, opened 2015-01-06
ensure result is returned from function

dereklwood:
log line was interfering with the return value of a function.  fixed.


wolf31o2:
:+1: LGTM

PR #880, opened 2015-01-06
Move coopr-provisioner to a submodule

wolf31o2:
Having `coopr-provisioner` in a submodule makes it easier to do Ruby-specific testing. It allows the project to be managed as a Ruby project, rather than as a Ruby component of a Java project.


duddun:
LGTM

PR #882, opened 2015-01-06
Only run sudoers code when /etc/sudoers exists

wolf31o2:
There's no `/etc/sudoers` on CoreOS, so we have to make sure it exists before we try to sed it.


duddun:
LGTM

wolf31o2:
Closing in favor of caskdata/coopr-provisioner#3

PR #886, opened 2015-01-08
COOPR-632: Update CLI to support partial templates

jwang47:


jwang47:
Review was done, but there were merge conflicts which git automatically resolved. Will merge after build passes.

PR #887, opened 2015-01-08
COOPR-300: Move lease times away from milliseconds

jwang47:


jwang47:
Build appears to have passed, despite what the GitHub widget is saying.. https://travis-ci.org/caskdata/coopr/builds/46264549

wolf31o2:
Actually, the end-to-end UI test failed. It just doesn't abort the build due to how it's setup. However, the Maven tests for the Java passed, so I think it's still good. We just need a Java resource to code review.

coopr-server/src/main/java/co/cask/coopr/spec/template/LeaseDuration.java
```diff
@@ -130,4 +190,48 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hashCode(initial, max, step);
   }
+
+  /**
+   * Builder for {@link LeaseDuration}.
+   */
+  public static final class Builder {
```
albertshau, line 199: If we have a builder, the constructors should be private so that objects are only created by builders.  To be consistent with the other classes, should also have a static builder() method so that you can do `LeaseDuration.builder()` instead of newing it.

coopr-server/src/main/java/co/cask/coopr/spec/template/LeaseDuration.java
```diff
@@ -130,4 +190,48 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hashCode(initial, max, step);
   }
+
+  /**
+   * Builder for {@link LeaseDuration}.
```
albertshau, line 197: It's not immediately clear how we expect timestamps as strings to be formatted. Add a javadoc so it is clear.

wolf31o2:
I restarted the test and it passed this time. I suspect it was a transient browserstack issue.

coopr-server/src/main/java/co/cask/coopr/spec/template/LeaseDuration.java
```diff
@@ -38,6 +40,50 @@ public LeaseDuration(long initial, long max, long step) {
...
+  }
+
+  /**
+   * Returns a timestamp in milliseconds.
+   *
+   * @param base The string argument user provided.
+   * @return Timestamp in milliseconds
+   */
+  private static long getTimestamp(String base) {
+    if (base.equals("min") || base.equals("0")) {
```
albertshau: initial == 0 means there is no lease, and max == 0 means there is no max.  Making 0 have special meaning is confusing, but I don't think we should change the behavior.  I think we should remove the "min" and "max" keywords.  Something like:

```
"leaseDuration": {
  "initial": "min",
  "max": "max",
  "step": "max"
}
```

would be very confusing for example.

coopr-server/src/main/java/co/cask/coopr/spec/template/LeaseDuration.java
```diff
@@ -73,7 +119,21 @@ public long getStep() {
    * @param requestedInitialLease Requested initial lease.
    * @return The smaller of the leases.
    * @throws InvalidClusterException if the requested lease is larger than the allowed initial lease, or if it is
-   *                                  less than negative one.
+   *                                 less than negative one.
+   */
+  public long calcInitialLease(String requestedInitialLease) throws InvalidClusterException {
```
albertshau, line 126: Is this used anywhere? We should not be using strings internally.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/LeaseDurationCodec.java
```diff
@@ -34,11 +35,44 @@ public LeaseDuration deserialize(JsonElement json, Type typeOfT, JsonDeserializa
     throws JsonParseException {
     JsonObject jsonObj = json.getAsJsonObject();
 
-    Long initial = context.deserialize(jsonObj.get("initial"), Long.class);
-    Long max = context.deserialize(jsonObj.get("max"), Long.class);
-    Long step = context.deserialize(jsonObj.get("step"), Long.class);
+    if (jsonObj.has("initial") && jsonObj.get("initial").isJsonPrimitive() &&
```
albertshau: it seems like this could be simplified to always send initial, max, and step as strings since the builder handles the conversion anyway.

coopr-server/src/main/java/co/cask/coopr/spec/template/LeaseDuration.java
```diff
@@ -38,6 +40,50 @@ public LeaseDuration(long initial, long max, long step) {
...
+          return TimeUnit.MINUTES.toMillis(offset);
+        case 'h':
+          return TimeUnit.HOURS.toMillis(offset);
+        case 'd':
+          return TimeUnit.DAYS.toMillis(offset);
+        default:
+          throw new RuntimeException("Unsupported relative time format: " + type);
+      }
+    } catch (NumberFormatException e) {
+      throw new RuntimeException("Invalid number value: " + base + ". Reason: " + e.getMessage());
```
albertshau: Do the http handlers catch this and return a 400?  Don't remember, but they might only look for IllegalArgumentExceptions so we might want to change the exception here.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/LeaseDurationCodec.java
albertshau: though it is possible they are null if they were not specified in the json.  If they are not there, I think it makes sense to default the values to 0, which could be done in the builder.

albertshau:
a few comments about the api and default values.

coopr-server/src/main/java/co/cask/coopr/codec/json/current/LeaseDurationCodec.java
```diff
@@ -34,11 +34,12 @@ public LeaseDuration deserialize(JsonElement json, Type typeOfT, JsonDeserializa
     throws JsonParseException {
     JsonObject jsonObj = json.getAsJsonObject();
 
-    Long initial = context.deserialize(jsonObj.get("initial"), Long.class);
-    Long max = context.deserialize(jsonObj.get("max"), Long.class);
-    Long step = context.deserialize(jsonObj.get("step"), Long.class);
+    LeaseDuration.Builder builder = LeaseDuration.builder();
+    builder.setInitial(jsonObj.getAsJsonPrimitive("initial").getAsString());
```
albertshau: I think, but am not sure, that this fails with a NPE if there is no "initial" key, which is why `context.deserialize(jsonObj.get("initial"), String.class` is used in other places. We should add a test case where one is missing to see how it behaves and make sure we handle it correctly.

albertshau:
one comment, then lgtm

PR #888, opened 2015-01-08
Always register provisioner plugins on standalone start

wolf31o2:
This allows for adding plugins to a standalone instance without requiring the user to manually perform a registration or delete their data.

### DIFF
--- a/coopr-standalone/bin/coopr.sh
+++ b/coopr-standalone/bin/coopr.sh
@@ -215,9 +215,6 @@ load_defaults () {
   echo "Loading default configuration..."
   ${COOPR_HOME}/server/templates/bin/load-templates.sh && touch ${COOPR_DATA_DIR}/.load_defaults || die "Couldn't upload templates"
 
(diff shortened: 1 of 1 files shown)


coopr-standalone/bin/coopr.sh
```diff
@@ -345,7 +342,7 @@ greeting () {
 
 stop () { provisioner stop; server stop; ui stop; }
 
-start () { server start && ui start && provisioner start && load_defaults && greeting; }
+start () { server start && ui start && provisioner register && provisioner start && load_defaults && greeting; }
```
duddun, line 345: does plugin registration depend on server start or is this just a cleaner way to start all of these?

duddun:
LGTM

coopr-standalone/bin/coopr.sh
wolf31o2, line 345: It depends on server start. Everything depends on server start.

PR #889, opened 2015-01-08
Standalone in debug

wolf31o2:
This sets debug logging for `coopr-standalone` by default, which aides Coopr development.


duddun:
:+1:

PR #890, opened 2015-01-08
COOPR-659: Update template APIs to support extending templates from mandatory partials

jwang47:


albertshau:
It makes more sense to me if mandatory is just a property of a partial instead of its own entity.  What was the reasoning for making it separate?

Also, it seems like there should be some enforcement that only one version of a partial can be mandatory.

albertshau:
Is this high priority? I would consider dropping it for now and revisiting it.

PR #891, opened 2015-01-08
COOPR-372: On plugin resource upload, check if the resource already exists

jwang47:


jwang47:
We'll have to change this PR to have the server compute the checksum and store it along with the entity, so that it can use the stored checksum when checking if the resource already exists.

PR #892, opened 2015-01-08
COOPR-652: Set connection close on calls to /status

jwang47:


wolf31o2:
Assuming Travis passes, it LGTM... :+1:

PR #855, opened 2014-12-17
Coopr 589:Document SSL/TLS usage for coopr-standalone

artemmakogon:
@jwang47, @wolf31o2  please review

PR #893, opened 2015-01-09
COOPR-589 Document TLS/SSL usage (originally #855)

wolf31o2:
This documents the current behavior.


coopr-docs/docs/source/installation/ui-config.rst
```diff
@@ -0,0 +1,45 @@
+..
+   Copyright © 2012-2014 Cask Data, Inc.
```
duddun: s/2014/2015/

PR #894, opened 2015-01-10
Add .dockerignore

wolf31o2:
Reducing the build context reduces the amount of data that has to be sent to the Docker daemon during build, speeding up the builds.


jwang47:
LGTM

PR #895, opened 2015-01-13
Use CDAP 2.6.0 rather than 2.6.0-SNAPSHOT

jwang47:


wolf31o2:
:+1: That was too easy!

PR #898, opened 2015-01-21
COOPR-665: API endpoints should only return a single version

jwang47:


coopr-server/src/main/java/co/cask/coopr/store/entity/BaseSQLEntityStoreView.java
```diff
@@ -172,21 +173,24 @@ protected PreparedStatement getSelectStatement(Connection conn, EntityType entit
...
-    // immune to sql injection since everything is an enum or constant
-    StringBuilder queryStr = new StringBuilder();
-    queryStr.append("SELECT ");
-    queryStr.append(entityTypeId);
-    queryStr.append(" FROM ");
-    queryStr.append(entityTypeId);
-    queryStr.append("s WHERE tenant_id=?");
+  private PreparedStatement getSelectAllLatestStatement(Connection conn, EntityType entityType) throws SQLException {
+    StringBuilder queryBuilder = new StringBuilder();
+    queryBuilder.append("SELECT ").append("t.").append(entityType.getBlobColumn()).append(", t.name, t.version");
```
albertshau: first 2 appends can be combined

coopr-server/src/main/java/co/cask/coopr/store/entity/BaseSQLEntityStoreView.java
```diff
@@ -172,21 +173,24 @@ protected PreparedStatement getSelectStatement(Connection conn, EntityType entit
...
-  private PreparedStatement getSelectAllStatement(Connection conn, EntityType entityType) throws SQLException {
-    String entityTypeId = entityType.getId();
-    // immune to sql injection since everything is an enum or constant
-    StringBuilder queryStr = new StringBuilder();
-    queryStr.append("SELECT ");
-    queryStr.append(entityTypeId);
-    queryStr.append(" FROM ");
-    queryStr.append(entityTypeId);
-    queryStr.append("s WHERE tenant_id=?");
+  private PreparedStatement getSelectAllLatestStatement(Connection conn, EntityType entityType) throws SQLException {
```
albertshau, line 176: this works right now, but is not efficient. It will also break if the "latest" version is not the max version. For example, if we want to support rolling back to an earlier version, we might want to tag the earlier version as "latest", but keep the higher version around.  Can we open a jira and reference it in a comment here to improve this logic.

albertshau:
small comments, then lgtm

jwang47:
Looks like maven test passed according to Travis. Merging now.

PR #901, opened 2015-01-27
COOPR-585 Better Coopr Server memory allocation

wolf31o2:
Defaulting this is a better solution than detecting if `/opt/coopr` exists.


duddun:
:+1:

PR #902, opened 2015-01-29
Document git submodule commands

wolf31o2:


rns4731:
LGTM

duddun:
Why did you remove the comments in the README?
They seemed helpful.

wolf31o2:
The commands are safe to always run, so there's no need to include `Git 101` documentation.

PR #904, opened 2015-01-29
COOPR-689 Mock provisioner API port

wolf31o2:


duddun:
Quick question: Why does COOPR_SERVER_URI point to localhost, rather than a variable that would point to whatever COOPR_SERVER_URI may be?  Does it have to be localhost?  If so, LGTM. :)

wolf31o2:
That variable is COOPR_SERVER_URI. The problem is that we're not connecting to COOPR_SERVER_URI, rather to the Coopr Server's tasks port. The only way to get this is via `server.tasks.port` in `coopr-site.xml`, but parsing that port requires a good bit of effort. The Mock provisioner is a debugging tool, and this is only being set in standalone's `coopr.sh`, so I think it's fine to leave as is, for now.

PR #906, opened 2015-01-29
Fix MockProvisionerMain default port for coopr-server

jwang47:
We separated /v2/\* into /v2/provisioners as an internal handler and the rest as external handlers.


wolf31o2:
:+1: LGTM

PR #907, opened 2015-01-30
COOPR-691: Fix coopr-cli interaction with remote coopr-server instances

jwang47:
- coopr-cli now works against remote coopr-servers
- coopr-cli now starts even if coopr-server isn't running on localhost with port 55054

### DIFF
--- a/coopr-cli/src/main/java/co/cask/coopr/shell/CLIConfig.java
+++ b/coopr-cli/src/main/java/co/cask/coopr/shell/CLIConfig.java
@@ -20,18 +20,19 @@ import co.cask.cdap.security.authentication.client.AccessToken;
 import co.cask.cdap.security.authentication.client.AuthenticationClient;
 import co.cask.cdap.security.authentication.client.Credential;
 import co.cask.cdap.security.authentication.client.basic.BasicAuthenticationClient;
+import co.cask.coopr.client.rest.RestClientConnectionConfig;
 import co.cask.coopr.client.rest.RestClientManager;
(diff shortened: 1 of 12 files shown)


coopr-cli/src/main/java/co/cask/coopr/shell/CLIConfig.java
```diff
@@ -136,16 +146,21 @@ public void setConnection(String host, int port, boolean ssl, String userId, Str
...
+    RestClientConnectionConfig.Builder builder = RestClientConnectionConfig.builder(host, port);
     builder.ssl(ssl);
-    accessToken = getAccessToken(host, port, ssl);
-    builder.accessToken(Suppliers.ofInstance(accessToken));
+
+    try {
+      accessToken = getAccessToken(host, port, ssl);
+      builder.accessToken(Suppliers.ofInstance(accessToken));
+    } catch (IOException e) {
+      // IGNORE
```
albertshau: when is this thrown? if there is no access token? Do we need to log anything? At least have a more descriptive comment

albertshau:
:+1:

coopr-cli/src/main/java/co/cask/coopr/shell/CLIConfig.java
jwang47: If there's a connection issue then this will be thrown. The reason for ignoring this was because I didn't want to exit if there's an issue connecting to the default connection, but I've refactored this so that I can throw IOException from here without affect CLI startup.

PR #908, opened 2015-02-04
input boxes for lease extension

rns4731:
replace slider with input boxes
https://issues.cask.co/browse/COOPR-686?filter=-1

### DIFF
--- a/coopr-ui/app/features/clusters/controllers/detail-ctrl.js
+++ b/coopr-ui/app/features/clusters/controllers/detail-ctrl.js
@@ -3,7 +3,7 @@
  */
 
 angular.module(PKG.name+'.feature.clusters').controller('ClusterDetailCtrl',
-function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, myApi, $timeout, moment) {
+function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, myApi, $timeout, moment, myHelpers) {
 
   CrudFormBase.apply($scope);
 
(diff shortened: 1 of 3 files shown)


coopr-ui/app/features/clusters/controllers/detail-ctrl.js
```diff
@@ -78,18 +78,35 @@ function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, m
       maxDate = moment.min(maxDate, moment().add(ld.step, 'ms'));
     }
     $scope.leaseMaxMs = maxDate.valueOf() - Date.now();
+    $scope.leaseMaxHumanized = myHelpers.millisecondsToString($scope.leaseMaxMs);
```
gaarf: `leaseMaxHumanized` is not used in template, could we just construct it on the fly in `showLeaseAlert()` ?

coopr-ui/app/services/helpers.js
```diff
@@ -34,7 +34,28 @@ angular.module(PKG.name+'.services').value('myHelpers', {
...
     }
     return total * 1000;
+  },
+
+  /**
+   * Get human readable string from milliseconds.
+   * @return {String} readable millisecond display.
+   */
+  millisecondsToString: function (milliseconds) {
+    var parsedTime = this.parseMilliseconds(milliseconds);
```
gaarf: not sure we can guarantee the method will be called with the correct context for `this` to work.

coopr-ui/app/services/helpers.js
```diff
@@ -34,7 +34,28 @@ angular.module(PKG.name+'.services').value('myHelpers', {
...
+    var readableString = [];
+    if (parsedTime.hasOwnProperty('days')) {
+      readableString.push(parsedTime.days + ' days');
+    }
+    if (parsedTime.hasOwnProperty('hours')) {
+      readableString.push(parsedTime.hours + ' hours');
+    }
+    if (parsedTime.hasOwnProperty('minutes')) {
```
gaarf: all 3 of the `hasOwnProperty` checks will always yield true.

coopr-ui/app/features/clusters/controllers/detail-ctrl.js
```diff
@@ -137,6 +153,13 @@ function ($scope, MYSERVICEPICKER_EVENT, CrudFormBase, $state, $modal, $alert, m
 
   /* ----------------------------------------------------------------------- */
 
+  function showLeaseAlert() {
+    $alert({
+      title: 'Lease extension maxed: ',
+      content: 'maximum lease is ' + $scope.leaseMaxHumanized,
```
gaarf: build the content string on the fly, something like `moment($scope.leaseMaxMs)` maybe ?

rns4731: thought leaseMaxHumanized never changes so we should construct it once, added a variable instead of declaring on scope. Moment's humanize() produces something like "maximum lease is 3 months" which is not specific enough.

gaarf:
LGTM :+1:

PR #909, opened 2015-02-05
register must wait for server during initial startup

dereklwood:
fixing standalone initial startup bug introduced in #888


wolf31o2:
:+1: LGTM

PR #911, opened 2015-02-09
COOPR-681 CLI should provide usage when given invalid command

jwang47:
```
coopr (http://localhost:55054)> invalidcommand
Invalid command: 'invalidcommand' - Enter 'help' for a list of commands
```


gokulavasan:
LGTM :+1:

PR #914, opened 2015-02-13
update standalone instructions to include deep_merge gem

dereklwood:
- [x] standalone README update needed for https://github.com/caskdata/coopr-provisioner/pull/32
- [ ] For a subsequent PR, I think we can actually just use bundler now for standalone.


wolf31o2:
:+1: LGTM

PR #916, opened 2015-02-14
Update coopr-templates/coopr-provisioner submodules

wolf31o2:
Update to the latest `release/0.9.9` branches on each.


dereklwood:
:+1:

PR #917, opened 2015-02-16
fix plugin dir path for new provisioner layout

dereklwood:
updating path to plugin directory after provisioner layout changes

this PR is to develop, since it needs to be fixed there.  I will then get it into release/0.9.9


duddun:
LGTM

wolf31o2:
:+1: LGTM

PR #922, opened 2015-02-19
use provided providerFields in my-provider-fields directive

gaarf:
May fix https://issues.cask.co/browse/COOPR-656

PR #923, opened 2015-02-19
use provided providerFields in my-provider-fields directive

gaarf:
same as #922 but against release/0.9.9 branch


dereklwood:
:+1: pending travis

PR #924, opened 2015-02-20
Feature/r0.9.9 fix links

awholegunch:
Fixes minor issues in the Coopr docs:
- Consolidated two images with all other images;
- Adjusted formatting; 
- Removed a footer file that was no longer being used; and
- Fixed two broken links.


wolf31o2:
:+1: LGTM...

PR #925, opened 2015-02-20
COOPR-703 Fix upgrade script to work on MySQL

jwang47:
Upgrade script was previously tested against MS SQL, but we need MySQL.


wolf31o2:
:+1: LGTM

wolf31o2:
Err.. could you do this against `release/0.9.9` also?

jwang47:
Oops.. will do.

PR #927, opened 2015-02-20
COOPR-593: Added UpgradeTo0_9_9 script to handle changes introduced by node usage APIs

jwang47:
This adds an upgrade script that upgrades a database from 0.9.8 to 0.9.9 for the changes in PR https://github.com/caskdata/coopr/pull/757.

### DIFF
--- a/coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
+++ b/coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
@@ -1,0 +1,216 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package co.cask.coopr.upgrade;
+
+import co.cask.coopr.cluster.Cluster;
+import co.cask.coopr.common.conf.Configuration;
+import co.cask.coopr.common.conf.Constants;
+import co.cask.coopr.spec.ProvisionerAction;
+import co.cask.coopr.store.DBConnectionPool;
+import co.cask.coopr.store.cluster.ClusterStore;
+import co.cask.coopr.store.cluster.SQLClusterStoreService;
+import co.cask.coopr.store.guice.StoreModule;
+import com.google.common.base.Charsets;
+import com.google.common.collect.Maps;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.inject.AbstractModule;
+import com.google.inject.Guice;
+import com.google.inject.Inject;
+import com.google.inject.Injector;
+import org.apache.commons.cli.BasicParser;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.OptionGroup;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Upgrades database to work with Coopr 0.9.9.
+ */
+public class UpgradeTo0_9_9 {
+
+  private static final Logger LOG = LoggerFactory.getLogger(UpgradeTo0_9_9.class);
+  private static final Gson GSON = new Gson();
+
+  private final DBConnectionPool dbConnectionPool;
+  private final ClusterStore clusterStore;
+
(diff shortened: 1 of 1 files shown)


jwang47:
Will also push this to release/0.9.9 once review is complete.

coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
```diff
@@ -0,0 +1,233 @@
...
+      CommandLine command = parser.parse(options, args);
+      String jdbcConnectionString = command.getOptionValue("c");
+      String jdbcDriver = command.getOptionValue("d", Driver.class.getName());
+      String dbMaxActiveConnections = command.getOptionValue("n", "100");
+      String dbUser = command.getOptionValue("u", "");
+      String dbPass = command.getOptionValue("p", "");
+      long batchSize = Long.parseLong(command.getOptionValue("b", "25000"));
+
+      final Configuration configuration = Configuration.create();
+      configuration.set(Constants.JDBC_CONNECTION_STRING, jdbcConnectionString);
```
albertshau: i think its fine to force people to use the same conf as the server instead of taking command line args. That way its guaranteed to connect to the correct db.

coopr-server/pom.xml
```diff
@@ -82,6 +82,14 @@
             </exclusion>
           </exclusions>
         </dependency>
+
+      <!-- REMOVE -->
+      <dependency>
```
albertshau: we don't need this right?

coopr-server/pom.xml
```diff
@@ -82,6 +82,14 @@
...
+        <groupId>mysql</groupId>
+        <artifactId>mysql-connector-java</artifactId>
+        <version>5.1.34</version>
+      </dependency>
```
jwang47: Not sure if we want to keep this, but it's helpful to make the `UpgradeTo0_9_9` script work against MySQL databases without having to bother with the classpath.

coopr-server/pom.xml
```diff
@@ -82,6 +82,14 @@
             </exclusion>
           </exclusions>
         </dependency>
+
+      <!-- REMOVE -->
+      <dependency>
+        <groupId>mysql</groupId>
+        <artifactId>mysql-connector-java</artifactId>
```
wolf31o2: No can do... not Apache 2 or compatible. This is GPL code.

coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
```diff
@@ -0,0 +1,233 @@
...
+    LOG.info("Denormalizing tasks table: type, cluster_template_name, user_id, and tenant_id columns");
+    List<Cluster> clusters = clusterStore.getAllClusters();
+    Map<Long, Cluster> clustersById = Maps.newHashMap();
+    for (Cluster cluster : clusters) {
+      clustersById.put(Long.parseLong(cluster.getId()), cluster);
+    }
+
+    Connection conn = dbConnectionPool.getConnection();
+    try {
+      PreparedStatement statement = conn.prepareStatement("SELECT * FROM tasks ORDER BY status_time ASC");
```
albertshau: dont need to order

coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
```diff
@@ -0,0 +1,233 @@
...
+      String dbUser = command.getOptionValue("u", "");
+      String dbPass = command.getOptionValue("p", "");
+      long batchSize = Long.parseLong(command.getOptionValue("b", "25000"));
+
+      final Configuration configuration = Configuration.create();
+      configuration.set(Constants.JDBC_CONNECTION_STRING, jdbcConnectionString);
+      configuration.set(Constants.JDBC_DRIVER, jdbcDriver);
+      configuration.set(Constants.DB_MAX_ACTIVE_CONNECTIONS, dbMaxActiveConnections);
+      configuration.set(Constants.DB_USER, dbUser);
+      configuration.set(Constants.DB_PASSWORD, dbPass);
```
wolf31o2: Can we have the tool read `coopr-site.xml` for this information, and maybe support a `--config-file` option to point to another file? I think this removes the need to the specific MySQL driver, as it'll use whatever's configured for Coopr (and already in the CLASSPATH)...

coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
```diff
@@ -0,0 +1,233 @@
...
+              ProvisionerAction type = ProvisionerAction.valueOf(taskName.toUpperCase());
+              updateStatement.setString(1, type.name());
+
+              Cluster cluster = clustersById.get(clusterId);
+              if (cluster != null) {
+                updateStatement.setString(2, cluster.getClusterTemplate().getName()); // 2: cluster_template_name
+                updateStatement.setString(3, cluster.getAccount().getUserId()); // 3: user_id
+                updateStatement.setString(4, cluster.getAccount().getTenantId()); // 4: tenant_id
+              } else {
+                updateStatement.setString(2, null); // 2: cluster_template_name
```
albertshau, line 124: this shouldn't be possible, should log an error

coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
albertshau: though i guess since it's already here it's fine to overwrite, but should be able to just use whats in the conf

coopr-server/pom.xml
albertshau: This assumes a specific version though. If they are upgrading, it means they should already have everything set up with the jdbc driver. So I think we should remove.

albertshau:
minor comments plus it should be able to work with their conf

wolf31o2:
It **has** to work with their conf/CLASSPATH, as only the Postgres Java connector has a compatible license. For MySQL/Oracle/etc. the user needs to supply the JARs, themselves. We cannot ship them.

albertshau:
:+1:

jwang47:
Maven build passed. Merging.

PR #930, opened 2015-02-25
COOPR-719 Limit select to correct tenant

wolf31o2:
The previous behavior did not limit the searches to a single tenant. As such, it's possible to return an empty set.

Before:

```
mysql> SELECT t.provider, t.name, t.version FROM providers t INNER JOIN(SELECT name, MAX(version) version FROM providers GROUP BY name) ss ON t.name = ss.name AND t.version = ss.version WHERE t.tenant_id='superadmin';
Empty set (0.00 sec)
```

After:

```
mysql> SELECT t.provider, t.name, t.version FROM providers t INNER JOIN(SELECT name, MAX(version) version FROM providers where tenant_id='superadmin' GROUP BY name) ss ON t.name = ss.name AND t.version = ss.version WHERE t.tenant_id='superadmin';

<snip>
8 rows in set (0.00 sec)
```


jwang47:
LGTM

PR #933, opened 2015-02-25
Fix UpgradeTo0_9_9 by refreshing DB connection

jwang47:
Connection was timing out, so instead obtain a new DB connection when appropriate.

### DIFF
--- a/coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
+++ b/coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
@@ -49,6 +49,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


coopr-server/src/main/java/co/cask/coopr/upgrade/UpgradeTo0_9_9.java
```diff
@@ -76,85 +77,124 @@ public void run(long batchSize) throws IOException, SQLException {
...
     try {
-      PreparedStatement statement = conn.prepareStatement("SELECT * FROM tasks");
+      int approxNumTasks = getTableRowCount(conn, "tasks");
+      tasks = new ArrayList<Task>(approxNumTasks);
+
+      // populate tasks
+      LOG.info("Fetching all tasks");
+      PreparedStatement selectAllTasks = conn.prepareStatement("SELECT * FROM tasks");
+      try {
+        ResultSet rs = selectAllTasks.executeQuery();
```
albertshau: we shouldn't load all tasks in memory, there can be quite a lot.  What you could do is get the max task id, then do tasks 0 - 1000, then 1001 - 2000, etc.

albertshau:
one comment

albertshau:
one comment

albertshau:
:+1:

PR #944, opened 2015-03-06
Rename User Profile to User Menu and remove profile information

wolf31o2:
The description was from the old UI, which has been removed.


duddun:
LGTM

PR #957, opened 2015-03-11
Cleanup of documentation from 0.9.9

awholegunch:
This fixes two documentation issues left over in the release. 

The first was the callback Javascript; it was formatted incorrectly in the conf.py, but fixed manually before being uploaded to the servers. The second is the build script option "all", which was failing due to a missing variable definition. Neither of these are required to be in the release, but should be in develop moving forward.


wolf31o2:
:+1: LGTM...

PR #958, opened 2015-06-17
update coopr-provisioner submodule ref

dereklwood:
updating submodule ref in develop to the latest.  previous ref was nearly a month prior to 0.99 and was broken.


dereklwood:
travis failures due to unrelated UI failures

PR #960, opened 2015-06-18
prefer 127.0.0.1 over localhost

dereklwood:
Standalone startup can sometimes hang on a mac if its trying to connect to ipv6.  Updating startup script to use `127.0.0.1` instead of `localhost`


duddun:
LGTM. Travis issue looks unrelated.

PR #961, opened 2015-09-04
fix server sql location

dereklwood:
correcting the location of the server `sql` directory in the installation guide


wolf31o2:
:+1: Ignore build failures.. not docs-related

PR #962, opened 2015-10-05
Update README.md with latest net-ssh changes

wolf31o2:
This is temporary until everything can be modified to use `bundle install` and `bundle exec` and tested with packaging.


dereklwood:
:+1: lgtm.  create https://issues.cask.co/browse/COOPR-781 for the permanent fix

PR #963, opened 2015-11-13
Remove Browserstack-based testing

wolf31o2:
This removes the external dependency on Browserstack from Coopr Server/UI testing via Travis.


dereklwood:
:+1: lgtm

PR #968, opened 2016-05-23
Start blank, rather than assume template named base exists

wolf31o2:
This prevents a user from having to remove services from the list and removes the requirement that a template named "base" exists on the server.


elfenheart:
👍  LGTM

PR #970, opened 2016-05-25
Update Dockerfile to support latest Coopr 0.9.9 repack

wolf31o2:
Attempting to rebuild the Docker image for caskdata/coopr-standalone resulted in a failure trying to install Gems (nokogiri failed due to missing zlib headers)... fixing that resulted in a Coopr Provisioner with missing Gems (net-scp)... Resolving that led me down a path of setting a couple Gem versions and also pulling in missing dependencies, like `deep_merge`...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends openjdk-6-jdk && \
     apt-get install -y \
       build-essential \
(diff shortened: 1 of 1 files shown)


awholegunch:
LGTM

wolf31o2:
Normally, I would wait for the tests, but they don't cover the Dockerfile, so I am going to go ahead and merge this.

PR #976, opened 2016-11-10
Default Coopr Server to listen on all interfaces

wolf31o2:
Have Coopr Server listen on all interfaces by default, rather than just localhost, as this is the most likely use case for the distributed install.


dereklwood:
👍 lgtm

PR #977, opened 2016-11-18
Codeclimate config

wolf31o2:
This was imported from codeclimate. The only difference is that I've excluded `coopr-integration-testing/resources` which includes external code, and we don't care about testing the code quality of external code.

screen shot 2016-11-18 at 1 34 25 pm


wolf31o2:
Build failure is due to upstream issues with Maven Central.

wolf31o2:
Since this is all new and only effects Codeclimate, I'm merging.

PR #979, opened 2017-01-10
Feature/upgrade sql

dereklwood:
Makes it possible to actually run https://github.com/caskdata/coopr/blob/develop/coopr-server/sql/upgrade-tables-pre0.9.9-to-0.9.9.sql after https://github.com/caskdata/coopr/blob/develop/coopr-server/sql/create-tables-mysql.sql

- [x] changes SQL statements to MODIFY altered columns instead of add as duplicate
- [x] removes unecessary table alterations already defined in create-tables-mysql.sql


wolf31o2 (approved):
👍 LGTM

PR #980, opened 2017-04-12
update node check to allow versions >1.0.0

dereklwood:
Quick fix to tolerate node.js versions > 1.0.0.  (at some point the better fix will be to import our version-checking functions from cdap)


wolf31o2 (approved):
Do we know that newer versions work? Code LGTM...

PR #981, opened 2017-11-10
Update Dockerfile for coopr-standalone

wolf31o2:
This includes several small changes to the Dockerfile to be able
to build a coopr-standalone Docker image.

- Move apt installs after COPY
- Use OpenJDK 7 instead of 6
- Use bundler gem to handle RubyGems installs
- Expose ports in a single command to reduce layers

Signed-off-by: Chris Gianelloni <wolf31o2@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,31 +20,7 @@
 #
 FROM ubuntu:12.04
 MAINTAINER Cask Data <ops@cask.co>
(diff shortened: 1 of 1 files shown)


wolf31o2:
Test failures are definitely unrelated.

mattwuenschel (approved):

